PR #7, opened 2018-08-06
moving Jenkinsfile to build repo

Kinlaw:


.travis/release_stable.sh
```diff
@@ -1,15 +1,13 @@
...
+cp ../dist .
 git config --global user.name $COMMIT_AUTHOR_USERNAME
 git config --global user.email $COMMIT_AUTHOR_EMAIL
-git remote add travis-build ${REPO}.git
-git fetch origin
-git branch -v -a
-git checkout -b stable
 git add .
 git commit -m 'Build Stable by Travis'
 git push --force --set-upstream travis-build stable:stable
```
ryelo: Will this work if you don't remote add?

ryelo: The reason we had it with the remote add first was due to the build repo not having a stable branch already. If not, it would create it. 

I guess we could just add them manually and then just `git push` instead of using the remote

Kinlaw: I kinda went full circle with this. I can no longer init the git repo in the dist folder because that puts the Jenkins file in the same folder as the build files and so when Jenkins goes to rsync everything, it also rsyncs the Jenkinsfile which was causing some kind of weirdness, which now that I think about it makes no sense. Maybe it was something else the whole time. I'll switch it back and see if I can reproduce it.

PR #18, opened 2018-10-02
adding modals on card variants

ryelo:


src/PresentationalComponents/Modal/ModalContent.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+const ModalContent = ({ app, variant, className, ...props }) => {
+
+    const modalClasses = classNames (
+        'ins-c-modal__content',
+        className
+    );
+
+    switch (variant) {
```
karelhala: This can be simplified with object literals
```JSX
return {
  notEntitled: <p className ={ modalClasses } { ...props }> { app } is not entitled </p>,
  notSetUp: <p className ={ modalClasses } { ...props }> { app } has not been set up </p>
}[variant]
```

ryelo: Fixed!

PR #31, opened 2019-03-26
Add QA release to auto travis

karelhala:
* All changes to master automatically pushed to qa and ci beta
* Added travis branch to beta branches as well so we have correct build for beta releases

### DIFF
--- a/.travis/custom_release.sh
+++ b/.travis/custom_release.sh
@@ -4,7 +4,7 @@ set -x
 
 # for now... push everywhere when master updates
 if [ "${TRAVIS_BRANCH}" = "master" ]; then
-    for env in ci
+    for env in ci qa
     do
         echo
         echo
@@ -15,7 +15,7 @@ if [ "${TRAVIS_BRANCH}" = "master" ]; then
 fi
 
 if [ "${TRAVIS_BRANCH}" = "master-stable" ]; then
-    for env in ci
+    for env in ci qa
     do
         echo
         echo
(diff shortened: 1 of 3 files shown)


.travis/custom_release.sh
```diff
@@ -24,3 +24,7 @@ if [ "${TRAVIS_BRANCH}" = "master-stable" ]; then
         .travis/release.sh "${env}-stable"
     done
 fi
+
+if [[ "${TRAVIS_BRANCH}" = "prod-beta" || "${TRAVIS_BRANCH}" = "prod-stable" ]]; then
```
iphands: Ouch we cant do this I think.... right?
The artifact needs to be rebuild for beta/stable. I think we want this split.

i.e. if we ship beta bits to prod stable ... the index.html will have /beta baked in right?

iphands: I might be wrong... though. Lemm know

karelhala: Not realy, if you look at [RedHatInsights/insights-dashboard/pull/31/files#diff-f737fedb5357468044db2dbe98eb0218R14](https://github.com/RedHatInsights/insights-dashboard/pull/31/files#diff-f737fedb5357468044db2dbe98eb0218R14) you notice that when building for `prod-stable` it will be stable build and then stable branch is pushed because of `TRAVIS_BRANCH` is set to `prod-stable` it's actually part of [RedHatInsights/insights-frontend-builder-common/blob/master/src/after_success.sh#L33](https://github.com/RedHatInsights/insights-frontend-builder-common/blob/master/src/after_success.sh#L33) script that is used by default.

iphands: Derrrr... its pushing beta and stable at the same time that goofs this not doing one or the other.
My bad :D thanks.

PR #33, opened 2019-04-04
Fixes/big resync with starter app

iphands:
@karelhala I started out on this "quest" because the dashboard in prod had /beta hardcoded.
I figured a sync with the starter app PR I worked on would be good.

I think everything worked out well... the only thing that is out of date is the PF next stuff.

I could not figure out what causes the artifact to build Stable vs Beta though, so I was unable to test the Beta case... whuups :D


package.json.new
```diff
@@ -0,0 +1,123 @@
+{
```
karelhala: I don't think this file was intentionally added.

src/App.js
```diff
@@ -7,15 +7,19 @@ import './App.scss';
...
 class App extends Component {
 
-    // init chrome
     componentDidMount () {
         insights.chrome.init();
         insights.chrome.identifyApp('dashboard');
     }
 
+    componentWillUnmount () {
+        this.appNav();
```
karelhala: There is no appNav and no buildNav so no need to call these 2 functions.

PR #42, opened 2019-09-04
Fetch correct number of CVEs with CVSS>=7

jiridostal:


jiridostal:
@karelhala can we pls review this? :-)

PR #49, opened 2020-02-17
add template cards & update release file

ryelo:
This should get us started on the template card for the new dashboard. We should make this robust enough so that at least teams could use the wrapper so we can add OUIA later.


AllenBW:
reviewing this as we speak... dot dot dot

PR #65, opened 2020-03-19
Populate data for Vulnerability card

katerinapat:
I removed a lot of stuff cause we don't need all the extra API calls anymore. 
Screenshot from 2020-03-25 14-21-32


katerinapat:
I can't edit the Reviewers list so please take a look on that: 

src/SmartComponents/Vulnerability/VulnerabilityCard.js
```diff
@@ -101,68 +92,52 @@ class VulnerabilityCard extends Component {
...
+                }
+            </TemplateCardBody>
+            <Divider/>
+            {vulnerabilitiesFetchStatus !== 'fulfilled' ? <Loading /> :
                 <TemplateCardBody isFilled={ false } isHorizontalLayout="true">
                     <NumberDescription
-                        data="NA"
+                        data={ vulnerabilities.recent_cves.last90days }
                         dataSize="md"
                         linkDescription="Last 90 days"
```
christiemolloy: Is it possible to populate this string for `linkDescription` with the last part of the string for `data` so that it's not written manually?

src/SmartComponents/Vulnerability/VulnerabilityCard.js
```diff
@@ -101,68 +92,52 @@ class VulnerabilityCard extends Component {
...
+                        link={ `${UI_BASE}/vulnerability/cves?publish_date=last90&show_irrelevant=true` }
                     />
                     <VerticalDivider/>
                     <NumberDescription
-                        data="NA"
+                        data={ vulnerabilities.recent_cves.last30days }
                         dataSize="md"
                         linkDescription="Last 30 days"
```
christiemolloy: same comment as above

christiemolloy:
This looks great @katerinapat ! I left two comments above.

After running your branch locally, I can confirm that the data is there for:
- total CVE's affecting the system
- all three CVSS ranges
-  the last 90 days, 30 days and 7 days
- the links for the time ranges 

Screen Shot 2020-03-20 at 9 22 25 AM

src/SmartComponents/Vulnerability/VulnerabilityCard.js
katerinapat: The response returns for example:
```
recent_cves = {
 last30days: 8,
 last7days: 0, 
 last90days: 21
}

```
I can grab the keys from the object (ex `last30days`) and pass it as a key for the translation 
`intl.formatMessage(messages.last30days)`  but that's unsafe because if for any reason the backend changes the response, the `react-intl` will throw an error that can't find the ID to format the message

src/Messages.js
```diff
@@ -62,6 +62,31 @@ export default defineMessages({
         description: 'Vulnerabilities - vulnerability total',
         defaultMessage: '{total}'
     },
+    last90days: {
```
AllenBW: rather than doing three of the same, what if we try something like this... would be useful for others possibly

    lastxdays: {
        id: 'lastxdays',
        description: 'Last x days label where x is a predefined value',
        defaultMessage: 'Last {days} days'
    },

src/SmartComponents/Vulnerability/VulnerabilityCard.js
```diff
@@ -7,88 +7,77 @@ import {
...
+    }
+
+    return (
+        <TemplateCard className='ins-c-card__vulnerability' data-ouia-safe={ vulnerabilitiesFetchStatus !== 'pending' }>
+            <TemplateCardHead>
+                <TemplateCardActions
+                    downloadReport="true"
+                    infoInlineMessage={ intl.formatMessage(messages.vulnerabilityLearnMore) }>
+                </TemplateCardActions>
+                <TemplateCardHeader title="Vulnerabilities"/>
```
AllenBW: can this be translated?

src/SmartComponents/Vulnerability/VulnerabilityCard.js
christiemolloy: Ok, thats a good point. I think its fine to hardcode it then

src/Messages.js
katerinapat: great catch :+1:  I will update it

kybaker:
Excellent, we do need to get the numbers to be link blue like on the latest recommendations, they should lick through to the vuln with the filtered list.

AllenBW:
@kybaker they will be, pr i merged earlier today is to the component

kybaker:
> @kybaker they will be, pr i merged earlier today is to the component
> 
> #75

Ack from me then

PR #68, opened 2020-03-20
Update stack chart template (remove stack chart from advisor)

christiemolloy:
Makes one stack chart template to be shared across Advisor and Vulnerability

**Don't merge until #65 is merged in** 

Screen Shot 2020-03-20 at 9 58 05 AM


AllenBW:
@christiemolloy any chance we can decouple this from #65  ? would be great to merge it sooner than later!

christiemolloy:
@AllenBW this shouldn't really interfere with #65 at all since it hardly touches the vulnerability file. I just didnt want to cause any conflicts for the dev, but I agree we should merge it right now

src/SmartComponents/Advisor/Advisor.js
```diff
@@ -26,6 +27,26 @@ const Advisor = ({ recStats, recStatsStatus, advisorFetchStatsRecs, advisorFetch
         advisorFetchIncidents();
     }, [advisorFetchIncidents, advisorFetchStatsRecs, advisorFetchStatsSystems]);
 
+    let chartData = [];
```
AllenBW: lets declare this using useState

src/SmartComponents/Advisor/Advisor.js
```diff
@@ -26,6 +27,26 @@ const Advisor = ({ recStats, recStatsStatus, advisorFetchStatsRecs, advisorFetch
...
+    if (recStatsStatus === 'fulfilled') {
+        let data = recStats.total_risk;
```
AllenBW: and do a useEffect on recStats where chartData is updated each time

christiemolloy:
@AllenBW thanks for the feedback. There will be a type error when clicking on the legend in the vulnerability card until the links for the vulnerability card are added in #65 . Could create logic to not add events to the legend, but that seems pointless seeing as we should have the link data very soon.

AllenBW:
travis passed, didn't report yet, merging now, asking for forgiveness later 🚜  https://travis-ci.org/github/RedHatInsights/insights-dashboard/builds/666490204

PR #70, opened 2020-03-20
add empty states for rbac

### LINKED ISSUE #56: Check RBAC permissions
We need to check RBAC permissions and then show an empty state if permissions are not found.

ryelo:
## What

Adds empty states when you don't have permissions. This utilizes the same pattern we use in other apps minus the button.

Closes #56

## Screenshots
Screen Shot 2020-03-20 at 1 19 00 PM

## Additional Notes
RBAC API doesn't allow doing a `GET` on multiple `/access/` so I have to do a big query and check to see if they're included. Not the best way imo, but we're limited. Will be updated in the future.

@kybaker


KendraMar:
messaging for no permissions looks good

ship it

kybaker:
ACK from me. Looks great.

src/App.js
```diff
@@ -1,17 +1,68 @@
...
+
+export const PermissionContext = createContext();
+
 const App = (props) => {
+
+    const [permissions, setPermissions] = useState({
+        customPolicies: false,
+        compliance: false,
+        drift: false,
+        insights: false,
```
AllenBW: wait this is still insights? shouldn't it be advisor ? (maybe insights is technically correct, butttttttt) 🤔

src/PresentationalComponents/DeniedState/DeniedState.js
```diff
@@ -0,0 +1,39 @@
...
+import { Card } from '@patternfly/react-core/dist/js/components/Card/Card';
+import { Title } from '@patternfly/react-core/dist/js/components/Title/Title';
+import {
+    EmptyState,
+    EmptyStateVariant,
+    EmptyStateIcon,
+    EmptyStateBody
+} from '@patternfly/react-core/dist/js/components/EmptyState';
+
+import { LockIcon } from '@patternfly/react-icons';
```
AllenBW: something like `import LockIcon from '@patternfly/react-icons/dist/js/icons/lock-icon';`

src/App.js
ryelo: I can change it to advisor, I wasn't sure since the actual permission is `insights:*:*`

AllenBW: ah ok, i wasn't sure what the permission was,  looks like yah gotta change it in this const

src/PresentationalComponents/DeniedState/DeniedState.js
```diff
@@ -0,0 +1,39 @@
...
+import { Card } from '@patternfly/react-core/dist/js/components/Card/Card';
+import { Title } from '@patternfly/react-core/dist/js/components/Title/Title';
+import {
+    EmptyState,
+    EmptyStateVariant,
+    EmptyStateIcon,
+    EmptyStateBody
+} from '@patternfly/react-core/dist/js/components/EmptyState';
+
+import { LockIcon } from '@patternfly/react-icons/dist/js/icons/lock-icon';
```
AllenBW: this throws 
```
Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: null.

Check the render method of `EmptyStateIcon`.

```

gotta do the import like this: `import LockIcon  from '@patternfly/react-icons/dist/js/icons/lock-icon';`

src/App.js
ryelo: Yeah I'll change it

ryelo:
@AllenBW I had a little issue with the merge conflicts, some changes got left out. Please check it all again 😄

PR #81, opened 2020-03-26
 Compliance - Nests pie charts and "more.." text in parent

AllenBW:
also auto orders imports, too much fun to not include

## Prerequisites

## What

Compliance card was broken so we fixed it

### Before
SUPAH BROKEN

### After
Screen Shot 2020-03-26 at 3 13 01 PM
Screen Shot 2020-03-26 at 3 12 43 PM

## Additional Info
Didn't fiddle with the actual code, obviously this needs to be done cuz title overrun 🙃 

cc @skarekrow


ryelo:
Merging for usability. I'll fix the design in another PR

PR #98, opened 2020-04-02
Update subscriptions card to fix indicactor and undefined/null states

### LINKED ISSUE #91: Fix progress bar in subscriptions utilized
The indicator isn't showing at all, based on this value being passed through: `value={ (productOne.percentage <= 100 && productOne.percentage) || 0 }`

Which means that any number over 100 will represent 0% for the indicator.

christiemolloy:
fixes #91 

## Prerequisites

- [X] Scope your styles to your individual card
- [X] You are using translations when necessary with proper plural/singular modifications
- [X] You are using the template card provided
- [X] Use functional components & hooks (please)
- [X] Attach screenshots (before and after)
- [X] Make sure that all text in blue is populated with the correct link. If you're unsure what the url should be, just ask!

## What

- The progress bar indicator was not showing because of the 0 value being passed through. Adjusted the value so that the percentage can just be passed through and even if a number over 100 is passed in it will always reflect 100%.
- Handles the state for undefined and null by completely removing the chart.

### Before

top one -->
Screen Shot 2020-03-31 at 2 35 09 PM
Screen Shot 2020-03-31 at 3 00 54 PM

### After
Screen Shot 2020-04-02 at 10 37 59 AM

Screen Shot 2020-04-02 at 10 22 37 AM

## Additional Info
[if needed]

## Code Review

## Design Review
@kybaker


christiemolloy:
@cdcabrera tagging you for review also

PR #107, opened 2020-04-07
Culls advisor incident call params

AllenBW:
param cleanup for the sake of clarity


AllenBW:
cc @dkuc

ryelo:
@dkuc is the fastest commenter in the west

PR #117, opened 2020-04-14
remove download report from compliance

ryelo:
## What

Removes download report from compliance


src/SmartComponents/Compliance/ComplianceCard.scss
```diff
@@ -40,4 +40,8 @@
     > .pf-l-stack .pf-l-stack__item .pf-l-split .pf-l-split__item {
       font-size: var(--pf-global--FontSize--sm);
     }
-  }
\ No newline at end of file
+  }
+
+.ins-c-card__compliance {
```
AllenBW, line 45: does this visually change the card?  should we apply this style to the template card (when sans actions?)

ryelo: Yeah, when there is no report button on this card or remediations, extra padding is applied. We could probably extract it somewhere. @christiemolloy and ideas on where to add it?

christiemolloy: umm id have to see it, want me to take this issue? @ryelo

ryelo: Sure thing. You'll notice this in both remediations and compliance if you remove the TemplateCardActions

christiemolloy: ok!

src/SmartComponents/Compliance/ComplianceCard.js
```diff
@@ -64,7 +63,6 @@ const ComplianceCard = ({ fetchCompliance, complianceFetchStatus, complianceSumm
             } : { 'data-ouia-safe': false } }
         >
             <TemplateCardHead>
```
christiemolloy, line 65: @ryelo i made changes and tried to push them up but its not working. Just remove the `<TemplateCardHead>` wrapper, and then also the styles you added

ryelo: Awesome! Will do!

PR #129, opened 2020-04-17
update empty state on top to use small text

ryelo:
## Prerequisites

## What

Changes the text in the top cards for the fail state

### Before
Screen Shot 2020-04-16 at 10 32 24 AM

### After
Screen Shot 2020-04-17 at 2 38 15 PM

## Code Review

## Design Review
@kybaker

### DIFF
--- a/src/PresentationalComponents/FailState/FailState.js
+++ b/src/PresentationalComponents/FailState/FailState.js
@@ -1,31 +1,37 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { Bullseye } from '@patternfly/react-core/dist/js/layouts/Bullseye/Bullseye';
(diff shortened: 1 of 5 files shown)


christiemolloy:
@kybaker should the empty state for system inventory be centered and the empty state for subscriptions be left-aligned?

Screen Shot 2020-04-17 at 2 40 54 PM

ryelo:
@christiemolloy those are actually two different states. Subscriptions will throw the one in that screenshot if we get a 403 back from the API, if we get a 404 or 500, it'll show the one currently in the Insights system inventory card. So I think if we do decide to left-align the top ones, it may look weird with the cards at the bottom all centered.

Screenshot if everything was to 500

Screen Shot 2020-04-17 at 2 53 06 PM

christiemolloy:
ok thanks for clarifying @ryelo

PR #142, opened 2020-04-28
make sure permissions exist before loading stuff

ryelo:
## What

When App.js loaded, if a user didn't have any permissions, there was no way you can map over an undefined object and check to see if it includes anything.

This skips that check and just uses the default `false` value if nothing is returned.


AllenBW:
goin to prod real fast, looks like something important

ryelo:
It's a weird edge case, but it's throwing logs to sentry and it's already been hit a few hundred times.

Issue #137, opened 2020-04-22
There's a hodge-podge of issues on the Insights dashboard


Ichimonji10:
There are several issues visible on both the [CRC insights dashboard](https://cloud.redhat.com/insights/dashboard) and [CRC beta insights dashboard](https://cloud.redhat.com/beta/insights/dashboard). For testing, I used Firefox 75.0 and Chromium 81.0.4044.113. I've attached screenshots at the bottom of this post.

Insights system inventory:

*   The header text is the same font as prose text.
*   The blue "i" icon doesn't show a tooltip when hovered or clicked.
*   The top link isn't underlined, but following links are.
*   Links span the width of the card.

Subscription watch utilization summary:

*   The header text is the same font as prose text.
*   The card is empty, with no explanation of why.

Patch:

*   The header text is the same font as prose text.
*   Hovering a mouse cursor over any section of the pie graph produces a popup that covers the pie graph and points off the pie graph.
*   The colors on the pie graph don't match the three colors on the legend.
*   The colors on the bar graph are hard to distinguish.

Advisor recommendations:

*   Hovering a mouse cursor over any section of the bar graph produces a popup that covers the pie graph and points to the wrong section of the pie graph, or doesn't point anywhere at all.
*   The colors on the bar graph are hard to distinguish.
*   Some links are underlined when moused over. Others are not.
*   Some links span the full width of the graph, even when no text is present.
*   The "incidents detected" text and the number next to it have differing vertical alignments.

Compliance:

*   The pie chart is nonsensical. There is no legend indicating what the sections mean. Hovering a cursor over one section generates a popup with the following text:

    > Standard System Security Profile for Red Hat Enterprise Linux 7: 38

    Hovering a cursor over the other section generates a popup with the following text:

    > empty: 100

    What's up with that?
*   The bottom of the card extends too far downwards.
*   There is a large gap between most of the card content and the "more compliance policies" link.

Vulnerability:

*   The "CVEs impacting your systems" text and the number next to it have differing vertical alignments.
*   Hovering a mouse cursor over any section of the bar graph produces a popup that covers the pie graph and points to the wrong section of the pie graph, or doesn't point anywhere at all.
*   The colors on the bar graph are hard to distinguish.
*   The meaning of the bottom portion is unclear. What does "1 last 90 days"
    mean? What does "0 last 30 days" mean? Is this for new vulnerabilities?

Remediations:

*   The top of the card is pushed down too far.
[...]

PR #144, opened 2020-05-04
Adjusts stack chart color, tooltip position

AllenBW:
addresses    https://github.com/RedHatInsights/insights-dashboard/issues/137#issuecomment-620217448 

addresses     https://github.com/RedHatInsights/insights-dashboard/issues/137#issuecomment-620229864

#### looks like
Screen Shot 2020-05-04 at 12 55 13 PM
Screen Shot 2020-05-04 at 12 55 01 PM


christiemolloy:
Your changes look good, I wonder why the advisor/vulnerability cards are growing outside of their parent width

Screen Shot 2020-05-04 at 3 38 59 PM

AllenBW:
the only way to get the tooltip on the top is to  kill `constrainToVisibleArea` as the chart doesn’t take up ah bunch of vertical space outside of the bar, the only place tooltips can go with constrainToVisibleArea enabled is on the chart

PR #148, opened 2020-05-12
Fix accessibility issues

christiemolloy:
## Prerequisites

- [X] Scope your styles to your individual card
- [X] You are using translations when necessary with proper plural/singular modifications
- [X] You are using the template card provided
- [X] Use functional components & hooks (please)
- [ ] Attach screenshots (before and after)
- [X] Make sure that all text in blue is populated with the correct link. If you're unsure what the url should be, just ask!

## What

Fixes accessibility issues in the dashboard
- All headings that are using `<p>` need a heading level of h2
- Fix the inline links where the number is not apart of the link so that it is read out incorrectly
- Added an HTML table so that the content inside of the charts is read out by the screen reader. We are currently making progress with this in PatternFly so for the time this is a solution that we can use.
- The links in the chart need to be able to be keyboard focusable, right now they are not usable and a screen reader cannot access them. Passed in an `<a>` tag for the links.
- The progress bar is reading the title but not the number % that is associated with the progress bar (this is a known issue in PF React - nothing we can do here)
- The title/descriptions associated with the charts needed to be updated

Screen Shot 2020-05-12 at 3 21 06 PM

## Code Review

## Design Review
@kybaker


src/SmartComponents/Advisor/Advisor.js
```diff
@@ -40,14 +40,13 @@ const Advisor = ({ recStats, recStatsStatus, advisorFetchStatsRecs, advisorFetch
...
-                `${UI_BASE}/advisor/recommendations?total_risk=${SEVERITY_MAP[risk]}&reports_shown=true&impacting=true&offset=0&limit=10`;
-        }
-    }];
+    const legendClick = chartData.map((data) => {
+        const risk = data.name.toLowerCase();
+        const link = `${UI_BASE}/advisor/recommendations?total_risk=${SEVERITY_MAP[risk]}&reports_shown=true&impacting=true&offset=0&limit=10`;
+        return link;
+    });
+
+    console.log('what is the legend click data' + legendClick);
```
AllenBW, line 49: eep

src/SmartComponents/Advisor/Advisor.js
```diff
@@ -66,7 +65,7 @@ const Advisor = ({ recStats, recStatsStatus, advisorFetchStatsRecs, advisorFetch
                 }
                 {recStatsStatus !== 'fulfilled' ? <Loading /> :
                     <StackChartTemplate
-                        ariaDesc="Advisor recommendations"
+                        ariaDesc="This chart represents the incidents detecting your system."
```
AllenBW, line 68: incidents a totally different thing, verbiage like... 'this chart shows recommendations effecting your system(s)'.

src/SmartComponents/Vulnerability/VulnerabilityCard.js
```diff
@@ -58,14 +58,11 @@ const VulnerabilityCard = ({ fetchVulnerabilities, vulnerabilities, vulnerabilit
...
-        target: 'labels',
-        mutation: (data) => {
-            const cvss = data.datum.name.split('CVSS ')[1];
-            window.location.href =
-                `${UI_BASE}/vulnerability/cves?cvss_filter=${CVSS_MAP[cvss]}`;
-        }
-    }];
+    const legendClick = chartData.map((data) => {
+        const cvss = data.name.split('CVSS ')[1];
+        const link = `${UI_BASE}/vulnerability/cves?cvss_filter=${CVSS_MAP[cvss]}`;
```
AllenBW, line 63: yah can just return this

src/SmartComponents/Advisor/Advisor.js
```diff
@@ -40,14 +40,13 @@ const Advisor = ({ recStats, recStatsStatus, advisorFetchStatsRecs, advisorFetch
...
-        target: 'labels',
-        mutation: (data) => {
-            const risk = data.datum.name.split(' ')[1].toLowerCase();
-            window.location.href =
-                `${UI_BASE}/advisor/recommendations?total_risk=${SEVERITY_MAP[risk]}&reports_shown=true&impacting=true&offset=0&limit=10`;
-        }
-    }];
+    const legendClick = chartData.map((data) => {
+        const risk = data.name.toLowerCase();
+        const link = `${UI_BASE}/advisor/recommendations?total_risk=${SEVERITY_MAP[risk]}&reports_shown=true&impacting=true&offset=0&limit=10`;
```
AllenBW, line 45: Can just return this line

src/ChartTemplates/StackChart/_StackChartTemplate.scss
```diff
@@ -7,3 +7,28 @@
...
 }
 
+.visually-hidden {
+    position: absolute !important;
+    top: -40px;
+    font-size: 10px;
+    height: 1px; 
+    width: 1px;
+    overflow: hidden;
+    clip: rect(1px 1px 1px 1px); /* IE6, IE7 */
```
AllenBW, line 17: gotta clean up the comments

src/ChartTemplates/StackChart/StackChartTemplate.js
```diff
@@ -52,23 +48,25 @@ export const StackChart = ({ ...props }) => {
...
+                }) }
+            </table>
             <ChartLegend
                 data={ legendData }
                 responsive={ false }
                 height={ props.legendHeight }
                 width={ props.legendWidth }
                 fontSize={ chartLegendFontSize }
                 className='pf-m-redhat-font'
-                style={ {
```
AllenBW, line 63: M not so sure about this change... it spawns two extra files and a buncha lines of code... does any behavior change?

christiemolloy: I spoke to Dan, and this change allows us to just pass in the StackChartLabel component and remove all the props for `style`, `labels` and `events`, because now the label is a button. This is the way we want to handle links inside of labels in charts from now on.

Do you think I should move the label file into this file to reduce the number of files?

christiemolloy:
updated and left a comment @AllenBW

src/ChartTemplates/StackChart/StackChartTemplate.js
AllenBW: Unless the component gonna be used in another file, yeah, keep everything in the one...

Totally down with killin the events, n thats the way it oughta be done now that we have the <a> there.. but the styles... is there a reason why we wanta remove them? especially when it takes 3x the code and a few `!importants` to get it to work?

christiemolloy: Great point. I just tried playing around with the style though, and I don't know how to change the hover/focus/active styles through the `styles` prop, do you?

Ideally it should look like `pf-c-button pf-m-link pf-m-inline` like the screenshot below when it is in the H/A/F state. If we can do that through the `styles` prop then that would be good, but otherwise I think we should have the stylesheet as it would be easier to maintain.

Screen Shot 2020-05-14 at 3 27 40 PM

AllenBW: OOO the difference that yer looking to fix being the color changes on hover? Yeah checkout the VictoryLegend documentation... specifically events... legend takes events n yah can do mutations n such https://formidable.com/open-source/victory/docs/victory-legend/#events

christiemolloy: Do you know if there is an event for `onHover` rather than `onClick` I've been playing around with the code but i can't seem to get the same implementation as just targeting the style sheet. ill keep working on it on Monday. I am a little hesitant though because for Accessibility reasons we will need the style change on `:hover` `:active` and `:focus` and I don't necessarily know that the events will cover all of these.

AllenBW: looks like `onMouseOver` is in fact an event yah can mutate on https://formidable.com/open-source/victory/guides/events/#victorysharedevents


some other reading here
https://formidable.com/open-source/victory/docs/common-container-props#events

christiemolloy: updated!

Screen Shot 2020-05-18 at 4 33 10 PM

AllenBW: Stellar!! looks great! last thing is to move that label piece back into the only file that uses it 👍  then we can merge thissssssss 🍕 🗡️

christiemolloy: updated :D

src/Messages.js
```diff
@@ -274,5 +274,10 @@ export default defineMessages({
         id: 'noSystemsLink',
         description: 'Getting started documentation',
         defaultMessage: 'Getting started documentation'
+    },
+    advisorChartDescription: {
+        id: 'advisorChartDescription',
+        description: 'Advisor chart description',
+        defaultMessage: 'This chart shows recommendations effecting your system}'
```
AllenBW, line 281: oooo dangling bracket! 🧼

PR #152, opened 2020-05-21
Fix wrong operation to compute link to 'more policies'

dLobatog:
## What

The operation to show "more policies" is wrong - if you have just 1 policy, it would do "3 - 1 > 1" and try to show you more links. The subtraction order is wrong, it should be the opposite "1 - 3 > 1" - in this case, it would not show you any links. 

### Before
Screenshot from 2020-05-21 18-11-05

### After
Screenshot from 2020-05-21 18-10-44

## Additional Info
[if needed]

## Code Review

## Design Review
@kybaker

### DIFF
--- a/src/SmartComponents/Compliance/ComplianceCard.js
+++ b/src/SmartComponents/Compliance/ComplianceCard.js
@@ -138,7 +138,7 @@ const ComplianceCard = ({ fetchCompliance, complianceFetchStatus, complianceSumm
                                         >
                                             {complianceFetchStatus === 'fulfilled' && Array.isArray(complianceSummary.data) &&
                                                 complianceSummary.data.length - 3 >= 1 &&
-                                                    `${3 - complianceSummary.data.length} more compliance reports`
+                                                    `${complianceSummary.data.length - 3} more compliance reports`
                                             }
                                         </Button>
                                     </div>
--- a/src/SmartComponents/Compliance/ComplianceCard.js
+++ b/src/SmartComponents/Compliance/ComplianceCard.js
@@ -137,7 +137,7 @@ const ComplianceCard = ({ fetchCompliance, complianceFetchStatus, complianceSumm
                                             isInline
                                         >
                                             {complianceFetchStatus === 'fulfilled' && Array.isArray(complianceSummary.data) &&
-                                                3 - complianceSummary.data.length >= 1 &&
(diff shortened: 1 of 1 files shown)


src/SmartComponents/Compliance/ComplianceCard.js
```diff
@@ -137,7 +137,7 @@ const ComplianceCard = ({ fetchCompliance, complianceFetchStatus, complianceSumm
                                             isInline
                                         >
                                             {complianceFetchStatus === 'fulfilled' && Array.isArray(complianceSummary.data) &&
-                                                3 - complianceSummary.data.length >= 1 &&
+                                                complianceSummary.data.length - 3 >= 1 &&
                                                     `${3 - complianceSummary.data.length} more compliance reports`
```
akofink, line 141: What is `3` anyway (besides a magic number)? We should define a descriptive constant.

akofink, line 141: Doesn't this also need to be updated?

```js
`${complianceSummary.data.length - 3} more compliance reports`
```

dLobatog: Right now the 3 is the max amount of policies displayed in the dashboard - we don't know how much longer we're going to keep this dashboard though

dLobatog: No, this one is accurate - if you have 5, 5-3 -> 2 more compliance reports. If you have 3 or less, it won't show up at all.

AllenBW:
@akofink @dLobatog we good for a merge here?

src/SmartComponents/Compliance/ComplianceCard.js
akofink: If you have 5, it's `3-5 -> -2`. What am I missing here?

akofink: This question wasn't because I didn't understand what it was but more to say: please define a constant, but if you're good with the magic number, then we can keep it.

dLobatog: I was reading the condition wrong - sorry about that, it's the same

dLobatog:
@AllenBW Now we are - text was still outdated, thanks!

PR #162, opened 2020-06-19
use per_page=1 for inventory to speed up calls

ryelo:
## What

Speed up inventory API calls by only using per_page=1, this will be updated in the future when `0` is supported. All we need is the metadata, no actual hosts.

## Screenshots
nothing changed


Glutexo:
Do you want `per_page=0` support?

ryelo:
If you want to add it, I'm happy to make another PR later to add that. I don't know how much more performant it would be just to get the metadata, but if you want to add it, we'll be happy to use it here!

PR #167, opened 2020-07-22
Fix sys-inv card tooltip, also fix text spacing

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-8080

as per @robcwilliams 's request, 'syste...' is now 'systems' plus that tooltip now works 

#### lookz like this
Screen Shot 2020-07-22 at 11 19 26 AM

Screen Shot 2020-07-22 at 11 16 48 AM


christiemolloy:
I noticed that the alignment here is looking different than it used to, as in the number used to be stacked on the text but now it is horizontally aligned.


Screen Shot 2020-07-22 at 11 20 03 AM

AllenBW:
OOOO @christiemolloy good point, didn't notice that change, couldn't find any others, wonder if this is a deal breaker @kybaker ?  (i suppose we could give each of the text items a size that would stack the text 🤔 )

kybaker:
@AllenBW The change to the inventory card looks great. For the vuln card we should stack the numbers that way as the numbers grow they don't break the layout.

AllenBW:
@kybaker @christiemolloy  ok so we're gonna allow flex direction to be explicitly specified on the numberdata component, previously that nasty button force it to column (destroying the tooltip) but now we can get column view n single line view without borking tooltip
Screen Shot 2020-07-22 at 6 00 21 PM

PR #182, opened 2020-09-09
Adds global filter support to dashboard and advisor card 🤗🍍🐞

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-9044

on filter change, we go into loading screen for the advisor widget then display the results

#### looks like

Screen Shot 2020-09-09 at 10 11 29 AM
Screen Shot 2020-09-09 at 10 11 12 AM
Screen Shot 2020-09-09 at 10 11 01 AM

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,19 @@
 import './App.scss';
 
-import React, { useEffect, createContext, useState } from 'react';
+import React, { createContext, useEffect, useState } from 'react';
 
-import { INVENTORY_TOTAL_FETCH_URL } from './AppConstants';
 import API from './Utilities/Api';
-
+import { INVENTORY_TOTAL_FETCH_URL } from './AppConstants';
+import PageLoading from './PresentationalComponents/PageLoading/PageLoading';
 import PropTypes from 'prop-types';
 import { Routes } from './Routes';
-
-import PageLoading from './PresentationalComponents/PageLoading/PageLoading';
+import { setSelectedTags } from './AppActions';
+import { useDispatch } from 'react-redux';
 
 export const PermissionContext = createContext();
 
 const App = (props) => {
-
+    const dispatch = useDispatch();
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -28,9 +28,23 @@ const App = (props) => {
...
-    async function initChrome () {
+    async function initChrome() {
         insights.chrome.init();
         insights.chrome.identifyApp('dashboard');
+
+        insights.chrome?.globalFilterScope?.('insights');
+        if (insights.chrome?.globalFilterScope) {
+            insights.chrome.on('GLOBAL_FILTER_UPDATE', ({ data }) => {
+                const selectedTags = insights.chrome?.mapGlobalFilter?.(data)?.filter(item => !item.includes('Workloads')) || undefined;
+                const encodedTags = selectedTags.map(tag => {
```
karelhala, line 39: Aren't you doing effectively the same thing twise? The mapGlobalFilter function returns an array of selected tags, you split them and encode them. Instead of this we should encode the tags in chrome function `mapGlobalFilter`

LI0131 (approved):
:godmode: 🏅  @AllenBW You may want to wait for a few more 👀 on this, but fwiw this looks good to me!

src/App.js
AllenBW: depends, some apis need it, maybe others dont? would be great if we didn't have to do this, but presently mapGlobalFilter doesn't encode the namespace/key/value

karelhala: I'm fine by encoding these things in chrome and if any app wants unencoded partials, they can always write their own mapper. Yeah, I figured that encoding full tag would look strange.

AllenBW: AWESOME! sweet! ok so if nothing else looks askew here, will merge this one, n remove the mapping encoder thingy once that shows up 🙇 🤗

PR #185, opened 2020-09-16
Updates advisor card to match new mocks

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1300

updates advisor card to support mock tooltip text is still missing (wont be merging till its finalized)

### where we're at so far 

Screen Shot 2020-09-17 at 11 30 30 AM
Screen Shot 2020-09-17 at 11 30 17 AM
Screen Shot 2020-09-17 at 11 30 06 AM


terezanovotna:
You're fast!

- can we have the divider between total risk and category right in the middle of the card? So it is aligned with the two cards above it
- for total risk - the labels should be all Capitalized (Important, Moderate, Low)

Overall: VERY NICE JOB!

src/SmartComponents/Advisor/Advisor.js
```diff
@@ -1,29 +1,59 @@
...
+    const colorScale = [
+        global_palette_cyan_100.value,
+        global_palette_cyan_200.value,
+        global_palette_cyan_300.value,
+        global_palette_cyan_400.value
+    ];
+
+    const pieLegendData = categoryData.map(item => ({ name: `${item.y} ${item.x}`, symbol: { fill: `${item.fill}`, type: 'square' } }));
+    const pieLegendClick = categoryData.map(({ value }) => `${UI_BASE}/advisor/recommendations?category=${value}&reports_shown=true&impacting=true&offset=0&limit=10${selectedTags.length && '&tags=' + selectedTags.join()}`);
+    const totalRiskUrl = (risk) => `${UI_BASE}/advisor/recommendations?total_risk=${risk}&reports_shown=true&impacting=true&offset=0&limit=10${selectedTags.length && '&tags=' + selectedTags.join()}`;
```
karelhala, line 47: This looks really complicated, might want to change it to. Shouldn't the selected tags be joined over `,`?
```suggestion
    const totalRiskUrl = (risk) => `${
      UI_BASE
    }/advisor/recommendations?total_risk=${
      risk
    }&reports_shown=true&impacting=true&offset=0&limit=10${
      selectedTags?.length && `&tags=${selectedTags.join()}`
    }`;
```

src/SmartComponents/Advisor/Advisor.js
```diff
@@ -1,29 +1,59 @@
...
+
+    const iconTooltip = text => <Tooltip
+        key={ `${text}` }
```
karelhala, line 50: No need for string interpolation
```suggestion
        key={text}
```

src/SmartComponents/Advisor/Advisor.js
```diff
@@ -32,55 +62,92 @@ const Advisor = ({ recStats, recStatsStatus, advisorFetchStatsRecs, advisorFetch
...
-    const legendClick = chartData.map((data) => {
-        const risk = data.name.toLowerCase();
-        // eslint-disable-next-line max-len
-        return `${UI_BASE}/advisor/recommendations?total_risk=${SEVERITY_MAP[risk]}&reports_shown=true&impacting=true&offset=0&limit=10${selectedTags.length && '&tags=' + selectedTags.join()}`;
-    });
-
     return <TemplateCard appName='Advisor' data-ouia-safe>
-        <TemplateCardHeader title='Advisor recommendations' />
+        {advisorIncidentsStatus === 'pending' ? <Loading /> : <TemplateCardHeader titleClassName='ins-m-red'
+            title={ `${intl.formatMessage(messages.incidents, { incidents: advisorIncidents?.meta?.count })}` }>
```
LI0131, line 97: Since this will still end up on the card when `advisorIncidentsStatus === 'rejected`, should provide a default value to `incidents` in the case that `advisorIncidents?.meta?.count` evaluates to `undefined`?

maybe something like `{ incidents: advisorIncidents?.meta?.count || 0 }` ?

src/SmartComponents/Advisor/Advisor.js
```diff
@@ -32,55 +62,92 @@ const Advisor = ({ recStats, recStatsStatus, advisorFetchStatsRecs, advisorFetch
...
-                        legendClick={ legendClick }
-                    />
-                }
-                {systemsStatsStatus !== 'fulfilled' ? <Loading /> :
-                    <Button component="a" href={ `${UI_BASE}${NEW_REC_URL}` } variant="link" isInline>
-                        {intl.formatMessage(messages.recsImpactingSystems, { totalRecs: recStats.total, systems: systemsStats.total })}
-                    </Button>}
+                <Flex>
+                    <FlexItem>
+                        <div><Title headingLevel='h2' size={ TitleSizes.lg }>
```
LI0131: Is the `div` here necessary?

LI0131 (changes requested):
This looks awesome @AllenBW :godmode: Love the new Pie Chart. Just one little thing but otherwise 💯 🏅

src/SmartComponents/Advisor/Advisor.js
AllenBW: yeah i was thinkin bout this believe the count will be zero not undefined, but doesn't hurt to add just in case

terezanovotna:
Very nice job!!

PR #188, opened 2020-09-23
Adds workload to app state, advisor card to consume/use

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1364

#### as yah can see, new attribute everywhere, but wont do much till advisor prod api supports it (also wont do much if the attribute doesn't exist)
Screen Shot 2020-09-23 at 9 36 49 AM

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,14 +1,14 @@
 import './App.scss';
 
 import React, { createContext, useEffect, useState } from 'react';
+import { batch, useDispatch } from 'react-redux';
+import { setSelectedTags, setWorkloads } from './AppActions';
 
 import API from './Utilities/Api';
 import { INVENTORY_TOTAL_FETCH_URL } from './AppConstants';
 import PageLoading from './PresentationalComponents/PageLoading/PageLoading';
 import PropTypes from 'prop-types';
 import { Routes } from './Routes';
-import { setSelectedTags } from './AppActions';
-import { useDispatch } from 'react-redux';
 
 export const PermissionContext = createContext();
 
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -35,6 +35,7 @@ const App = (props) => {
         insights.chrome?.globalFilterScope?.('insights');
         if (insights.chrome?.globalFilterScope) {
             insights.chrome.on('GLOBAL_FILTER_UPDATE', ({ data }) => {
+                data?.Workloads ? dispatch(setWorkloads(data.Workloads)) : dispatch(setWorkloads(undefined));
```
karelhala, line 38: Can be simplified with
```suggestion
                dispatch(setWorkloads(data?.Workloads))
```

AllenBW: oh slick yeah i'll change that

karelhala (approved):

src/App.js
```diff
@@ -35,6 +35,7 @@ const App = (props) => {
         insights.chrome?.globalFilterScope?.('insights');
         if (insights.chrome?.globalFilterScope) {
             insights.chrome.on('GLOBAL_FILTER_UPDATE', ({ data }) => {
+                dispatch(setWorkloads(data?.Workloads));
                 const selectedTags = insights.chrome?.mapGlobalFilter?.(data)?.filter(item => !item.includes('Workloads')) || undefined;
                 dispatch(setSelectedTags(selectedTags));
```
rvsia, line 40: what about to join these two dispatches into one?

rvsia (dismissed):
Just one small suggestion. Otherwise it looks good! :+1:

src/App.js
AllenBW: OOOOO how do we do that? 🤔 🥇

rvsia: I would just create a new action, that will combine both actions (so there will be only one re-rerender)

```jsx
dispatch(globalFilterChange(data?.workloads, selectedTags))
```

AllenBW: OOOOO i see `batch`

rvsia (approved):

karelhala (approved):

PR #194, opened 2020-10-05
Show empty state for inventory card if user has no access to inv

karelhala:
## What

If user doesn't have access to inventory error is thrown and error card is shown. We want to tell user more information, why that happend so let's show them that they have no access to inventory.

## Screenshots
**Wide screen**
Screenshot from 2020-10-05 15-59-25

**Mobile screen**
Screenshot from 2020-10-06 09-33-30

## Code Review

## Design Review
@kybaker @mmenestr

### DIFF
--- a/src/SmartComponents/SystemInventory/SystemInventoryCard.js
+++ b/src/SmartComponents/SystemInventory/SystemInventoryCard.js
@@ -96,10 +96,7 @@ const SystemInventoryCard = ({
                     </Fragment> : <NotAuthorized
                         showReturnButton={ false }
                         serviceName="Inventory"
-                        description={ <Fragment>
-                            <div>Your organization administrator must grant</div>
-                            <div>you inventory access to view your systems.</div>
-                        </Fragment> }
+                        description={ <div>{intl.formatMessage(messages.systemInventoryNoAccess)}</div> }
                     />
             }
         </TemplateCardBody>
--- a/src/SmartComponents/SystemInventory/SystemInventoryCard.js
+++ b/src/SmartComponents/SystemInventory/SystemInventoryCard.js
@@ -2,9 +2,10 @@ import './SystemInventoryCard.scss';
 
 import * as AppActions from '../../AppActions';
 
-import React, { useEffect } from 'react';
+import React, { useEffect, Fragment } from 'react';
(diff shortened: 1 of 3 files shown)


src/SmartComponents/SystemInventory/SystemInventoryCard.js
```diff
@@ -39,47 +45,62 @@ const SystemInventoryCard = ({
...
+                        }
+                        { inventoryTotalFetchStatus === 'rejected' &&
+                            <FailState appName='Inventory' isSmall/>
+                        }
+                    </Fragment> : <NotAuthorized
+                        showReturnButton={ false }
+                        serviceName="Inventory"
+                        description={ <Fragment>
+                            <div>Your organization administrator must grant</div>
+                            <div>you inventory access to view your systems.</div>
```
rvsia, line 101: What about to wrap them via using CSS?

karelhala: We want to break the sentence on specific place, at least that's on mocks.

rvsia: Yeah, I understand, I know that very well. :smiley:  But you should be able to setup the same width/max-width as in the mocks and the wrapping should correspond too.

rvsia: But it's definitely no blocker or whatever. (It can cause some troubles if we are going to translate this message _in the future_.)

AllenBW: I gotta ditto the translation thing, please throw the strings in Messages.js n use the `intl.formatMessage(` thing

AllenBW (changes requested):
Just the translations thing 🤗 👍

src/SmartComponents/SystemInventory/SystemInventoryCard.js
karelhala: Good poin! I've added the string in there and also styled it so it would fit the mocks and not used 2 divs.

rvsia (approved):

AllenBW (approved):
LUV IT, toats see that no access message being a lil more generic/ able to support each card in the future, but not today 🙈 🍍

AllenBW:
This work is to be pushed to prod-beta/stable at an unspecified time in the future (once those envs support rbac)

PR #196, opened 2020-10-08
Pass sap_system and tags to inventory enpoints

karelhala:
## What

Enable global filter tags and sap_system to be passed down to inventory enpoints. Use `generateFilter` from frontend components (it has been added recentrly) in order to properly style the filter for inventory to be `filter[system_profile][sap_syste]: true`. This function can later help us to calculate SAP_SIDs as well.

### Before
Screenshot from 2020-10-08 16-38-38
Screenshot from 2020-10-08 16-38-46

### After
Screenshot from 2020-10-08 16-37-03
Screenshot from 2020-10-08 16-37-37

## Additional Info
[if needed]

## Code Review

## Design Review
@kybaker

### DIFF
--- a/src/SmartComponents/SystemInventory/SystemInventoryCard.js
+++ b/src/SmartComponents/SystemInventory/SystemInventoryCard.js
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
 import './SystemInventoryCard.scss';
 
 import * as AppActions from '../../AppActions';
@@ -6,6 +7,7 @@ import React, { useEffect, Fragment } from 'react';
 import { TemplateCard, TemplateCardBody, TemplateCardHeader } from '../../PresentationalComponents/Template/TemplateCard';
 import { usePermissions } from '@redhat-cloud-services/frontend-components-utilities/files/RBACHook';
 import { NotAuthorized } from '@redhat-cloud-services/frontend-components/components/cjs/NotAuthorized';
+import { generateFilter } from '@redhat-cloud-services/frontend-components-utilities/files/helpers';
 import { Button } from '@patternfly/react-core/dist/js/components/Button/Button';
 import FailState from '../../PresentationalComponents/FailState/FailState';
 import { IconInline } from '../../PresentationalComponents/IconInline/IconInline';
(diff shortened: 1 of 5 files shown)


src/SmartComponents/SystemInventory/SystemInventoryCard.js
```diff
@@ -116,7 +138,20 @@ SystemInventoryCard.propTypes = {
     fetchInventoryTotal: PropTypes.func,
     inventoryTotalSummary: PropTypes.object,
     inventoryTotalFetchStatus: PropTypes.string,
-    intl: PropTypes.any
+    intl: PropTypes.any,
+    selectedTags: PropTypes.arrayOf(PropTypes.string),
+    workloads: PropTypes.oneOf([
```
AllenBW, line 143: this looks like a real specific shape, should we abstract it to a constants file so all apps can consume it?

src/SmartComponents/SystemInventory/SystemInventoryCard.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
```
AllenBW, line 1: we do this everywhere... any qualms with killing it globally in the app?

AllenBW (commented):
Just a few comments! overall love it, yer ah 👼 for doin <3

src/SmartComponents/SystemInventory/SystemInventoryCard.js
karelhala: Fine by me, you can disable this rule and have it to warning instead.

src/SmartComponents/SystemInventory/SystemInventoryCard.js
karelhala: Good point!

src/SmartComponents/SystemInventory/SystemInventoryCard.js
AllenBW: Heck yeah, i can do that in a followup, no need to add it here 🤗

src/SmartComponents/SystemInventory/SystemInventoryCard.js
karelhala: I've made it into complex prop type check so we have easier way of identifying what is wrong.

AllenBW (approved):
🥇

PR #198, opened 2020-10-12
Hide Advisor TemplateCardHeader if advisorIncidentsStatus is rejected

LI0131:
## Prerequisites

## What
Instead of showing `NaN` when an error is returned from the Advisor API, the Advisor card will now hide it's header and simply display the message about contacting Red Hat support.

### Before
Screen Shot 2020-10-12 at 1 39 50 PM

### After
Screen Shot 2020-10-12 at 1 27 07 PM

## Code Review

## Design Review
@kybaker


AllenBW (approved):
That'll do it 💯 🤗 👍

PR #199, opened 2020-10-12
Creates SAP zero state, temporarily hides inventory card rbac

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1400

and yes yes, temporarily disables inv card rbac, tired of cherry picking commits, will reenabled once it shows up in prod-beta or stable 🤷 

also yes yes, this will dupe the inventory fetch, not sure how to get around that, open to suggestions (meaning we'll do the first check to see if there are SAP systems when that workload is selected, if the are we'll load all the cards ahnew, including inventory one)

#### looks like 
Screen Shot 2020-10-13 at 11 11 26 AM


AllenBW:
also @LI0131  yer previous pr, the one hiding advisor card title will make it to prod-stable with when this one does 🤗 🙇

Robcwilliams:
PM Ack

terezanovotna:
Nice work!

The plus icon should be [grey](https://www.patternfly.org/v4/components/empty-state/#ws-page-title) or [here](https://www.patternfly.org/v4/components/empty-state/design-guidelines)

src/PresentationalComponents/Dashboard/Dashboard.js
```diff
@@ -21,13 +26,31 @@ const SubscriptionsUtilizedCard = lazy(() => import('../../SmartComponents/Subsc
...
 const RemediationsCard = lazy(() => import('../../SmartComponents/Remediations/RemediationsCard'));
 
-const Dashboard = () => {
-
+const Dashboard = ({ inventorySummary, workloads, fetchInventory, inventoryFetchStatus }) => {
     const permission = useContext(PermissionContext);
     const intl = useIntl();
+    const [supportsSap, setSupportsSap] = useState(true);
+
+    const fetchInventoryFn = useCallback(() => {
```
karelhala, line 34: Instead of fetching the full inventory information, you can query `/api/inventory/v1/system_profile/sap_system` and check the count for `value: true` result.

karelhala: But that requires a bit more code to be done (reducer, action, etc.)

karelhala (commented):
What if user selects specific tag that applies to 3 systems and none of them are SAP systems? I believe that this will yield in empty state result, which is not correct, right? IMO we could use the new enpoint `/api/inventory/v1/system_profile/sap_system` and store the number of systems under SAP without any filter applied.

src/PresentationalComponents/Dashboard/Dashboard.js
AllenBW: I left out the tags off the query for that very reason, but i agree, not super thrilled using the same endpoint, will checkout the other one yah mentioned! <3

terezanovotna:
Nice!

src/PresentationalComponents/Dashboard/Dashboard.js
AllenBW: cuz it also means i don't need the system count 😜

AllenBW: also means we can do it (the call) once n never again (till component is destroyed n recreated) rather than ever time the workplace is swapped

AllenBW:
Okie dokie, going with the new endpoint, but it only lives on ci for now, so this work is blocked till it shows up elsewhere

src/PresentationalComponents/Dashboard/Dashboard.js
```diff
@@ -21,13 +25,26 @@ const SubscriptionsUtilizedCard = lazy(() => import('../../SmartComponents/Subsc
...
+const Dashboard = ({ workloads }) => {
     const permission = useContext(PermissionContext);
     const intl = useIntl();
+    const [supportsSap, setSupportsSap] = useState(true);
+
+    useEffect(() => {
+        const fetchSapSystems = async () => {
+            try {
+                const response = await API.get(SAP_FETCH_URL);
+                setSupportsSap(response.data.results[1].count > 0);
```
karelhala, line 37: I wouldn't rely on this approach as the API could change order
```suggestion
                setSupportsSap(response.data.results?.find(({ value }) => value )?.count > 0);
```

karelhala (commented):
Just one minor improvement and we are golden!

src/PresentationalComponents/Dashboard/Dashboard.js
AllenBW: yah yah good call, THOUGH one would think changing up that in the api would be ah breaking change 🦹‍♀️🦹‍♀️🦹‍♀️ 🦹‍♀️ grrrrrr

karelhala (dismissed):

AllenBW:
This still wont work outside ci/qa beta, but @gmcculloug says that might change soon, so we'll merge it for now (also, without the aforementioned change, everything works the same, just... not as intended by this pr)

PR #200, opened 2020-10-13
Use new global filter mapper format and pass SID filter to inv

karelhala:
## What

This PR adds support for SID part of global filter and applies it for inventory card.

## Screenshots
Screenshot from 2020-10-13 13-57-25
Screenshot from 2020-10-13 13-57-38

## Code Review
@ryelo @AllenBW


AllenBW (approved):
LG2M

karelhala:
@AllenBW just be carefull, that this is using newer API and is not suitable for prod untill the API updates. I think we should push on the inv team to make it sooner rather than later.

PR #201, opened 2020-10-15
Patch card to support workloads/tagging/sids 🚤✅🥰

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1405

reverted app.js global filter handling to suite apps in place

#### looks LIIIIKEEEEEEEE this

Screen Shot 2020-10-15 at 12 29 16 PM
Screen Shot 2020-10-15 at 12 28 19 PM

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,7 +2,7 @@ import './App.scss';
 
 import React, { createContext, useEffect, useState } from 'react';
 import { batch, useDispatch } from 'react-redux';
-import { setSelectedTags, setWorkloads, setSIDs } from './AppActions';
+import { setSIDs, setSelectedTags, setWorkloads } from './AppActions';
 
 import API from './Utilities/Api';
 import { INVENTORY_TOTAL_FETCH_URL } from './AppConstants';
@@ -35,10 +35,11 @@ const App = (props) => {
         insights.chrome?.globalFilterScope?.('insights');
         if (insights.chrome?.globalFilterScope) {
             insights.chrome.on('GLOBAL_FILTER_UPDATE', ({ data }) => {
-                const [workloads, SID, selectedTags] = insights.chrome?.mapGlobalFilter?.(data, false, true);
+                let selectedTags = insights.chrome?.mapGlobalFilter?.(data)?.
+                    filter(item => !item.includes('Workloads')) || undefined;
                 batch(() => {
-                    dispatch(setWorkloads(workloads));
-                    dispatch(setSIDs(SID));
+                    dispatch(setWorkloads(data?.Workloads));
+                    dispatch(setSIDs(Object.entries(data?.['SAP ID (SID)'] || {}).filter(([, { isSelected }]) => isSelected).map(([key]) => key)));
                     dispatch(setSelectedTags(selectedTags));
                 });
             });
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -35,10 +35,11 @@ const App = (props) => {
         insights.chrome?.globalFilterScope?.('insights');
         if (insights.chrome?.globalFilterScope) {
             insights.chrome.on('GLOBAL_FILTER_UPDATE', ({ data }) => {
-                const [workloads, SID, selectedTags] = insights.chrome?.mapGlobalFilter?.(data, false, true);
```
karelhala, line 38: Can you revert this change? As the new API of `mapGlobalFilter` allows passing 3rd parameter to preformat the data for you? https://github.com/RedHatInsights/insights-chrome/blob/f0eb20df8671392fc0d3ef9aad6eb40b64659fb9/README.md#prefered-way-of-consuming-data-in-structured-format

AllenBW: this change breaks the dashboard workload selector (neither inventory or advisor cards work with workloads) furthermore, at least in prod-beta, SIDs dont exist yet n this change gobbles up tags, shows em as SIDS

AllenBW: also, regarding the link sent, is this work on prod beta? not seeing     // workloads = { SAP: { isSelected: true } } as the default value returned when using the mapGlobalFilter

karelhala: I forgot the API is still not merged to each environment... Yeah we can roll out with the older version of mapping global filter and I'll open new PR to enable the newer way once the API lands.

src/App.js
```diff
@@ -35,10 +35,11 @@ const App = (props) => {
...
+                let selectedTags = insights.chrome?.mapGlobalFilter?.(data)?.
+                filter(item =>  !item.includes('Workloads') || !item.includes('SAP ID (SID)')) || undefined;
                 batch(() => {
-                    dispatch(setWorkloads(workloads));
-                    dispatch(setSIDs(SID));
+                    dispatch(setWorkloads(data?.Workloads));
+                    dispatch(setSIDs(data['SAP ID (SID)']));
```
karelhala, line 42: Since we don't have the new way of mapping filter you might want to make this into an array
```JS
dispatch(setSIDs(Object.entries(data['SAP ID (SID)']).filter(([, { isSelected }]) => isSelected).map(([ key ]) => key)));
```

src/App.js
karelhala: The formatted API is just on ci/qa-beta and is blocked by API unfortunately.

karelhala (dismissed):

src/SmartComponents/Advisor/Advisor.js
```diff
@@ -60,12 +56,17 @@ const Advisor = ({ recStats, recStatsStatus, advisorFetchStatsRecs, advisorFetch
     const pieLegendData = categoryData.map(item => ({ name: `${item.y} ${item.x} `, symbol: { fill: `${item.fill} `, type: 'square' } }));
 
     useEffect(() => {
-        const options = { ...selectedTags.length && ({ tags: selectedTags.join() }), ...(workloads && workloadQueryBuilder(workloads)) };
+        const sapFilter = generateFilter({
```
LI0131, line 59: Would it be worthwhile to move `sapFilter` into the common utils file?

LI0131 (commented):
Looking good @AllenBW 😄

src/SmartComponents/Advisor/Advisor.js
AllenBW: hmmmm well every card that supports, will be using, and yah see we do the same thing in both patch and inv card... lemme give it a try n <3

karelhala (approved):

LI0131 (approved):

PR #203, opened 2020-10-21
Enable RBAC in inventory card, use new chrome API for global filter

karelhala:
## What

Re-enable RBAC for inventory card, reorder the render of inventory card so there is no flicker when inventory is being loaded. Also start using the new chrome API for splitting mapped global filter so we don't have to run it here.

### Before
Screenshot from 2020-10-05 15-59-25

### After
Screenshot from 2020-10-21 16-35-43

## Additional Info

Closes: RHCLOUD-9874

## Code Review
@ryelo @AllenBW


AllenBW (commented):
Just to confirm, prod-beta supports rbac and the updated mapGlobalFilter ?

karelhala:
@AllenBW yup, everything should be pushed all the way to prod-beta.

AllenBW (approved):
then i am A O K with this! 🤗 🙇

PR #204, opened 2020-10-21
Update messaging on cards that don't support global filter 

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1413

messaging appears whenever sid, workload or tag is selected (anything from the global filter)

##### updated mocks as per bu/ux requests
Screen Shot 2020-10-22 at 12 24 53 PM

#### and of course, still looks the same with all workloads
Screen Shot 2020-10-21 at 2 43 40 PM


karelhala:
@katerinapat @mkholjuraev do you know if vulnerability will support global filter in future? I know it can filter down systems, but what about number of CVEs?

karelhala (dismissed):

katerinapat:
@karelhala  currently we don't support filtering for this endpoint but we are planning to support it in the near future

kybaker:
Looks great! Ship it!

PR #210, opened 2020-11-16
Do not hide global filter trough CSS

karelhala:
## What

When user access dashboard, global filter is hidden all the time, this is caused by CSS rule, instead of hiding this trough CSS hide it trough chrome API.

## Code Review
@ryelo @AllenBW


AllenBW (approved):
🚢 🇮🇹 


:shipit:

PR #211, opened 2020-11-30
In the rare case inv is down, hide global filter on zero state 🙃🤤

AllenBW:
If we're displaying zero state, don't want global filter, period! 🙈


src/App.js
```diff
@@ -71,11 +71,11 @@ const App = (props) => {
                 setArePermissionReady(true);
             }
         );
-        const inventorySystems = await API.get(`${INVENTORY_TOTAL_FETCH_URL}`);
-        setHasSystems(inventorySystems.data.total > 0);
-        if (inventorySystems.data.total <= 0) {
-            insights.chrome.hideGlobalFilter();
-        }
+        await API.get(`${INVENTORY_TOTAL_FETCH_URL}`).then(response => {
```
karelhala, line 74: I don't really like mixing `await` and `then`. What do you think about wrapping this in `try/catch`?
```JS
try {
  const { total } = (await API.get(`${INVENTORY_TOTAL_FETCH_URL}`))?.data || { total: 0 };
  setHasSystems(total > 0);
  total === 0 && insights.chrome.hideGlobalFilter();
}  catch (e) {
  insights.chrome.hideGlobalFilter()
}
```

AllenBW: delicious! updated <3

Robcwilliams:
PM Ack +1

karelhala (approved):

PR #212, opened 2020-12-07
fix(compliance): RHICOMPL-1181 Group results by policy

akofink:
## Prerequisites

## What

Related to recent feature COMP-E-152, compliance reports should be
grouped by policy. This change should fix the dashboard to reflect the
results shown in the new compliance reports page.

In addition, unsupported reports are no longer counted in the compliant or total (renamed to test_result_count) count. For now, those hosts are not shown in the dashboard at all.

### Before
image

### After
image

## Additional Info
Requires https://github.com/RedHatInsights/compliance-backend/pull/677

## Code Review

## Design Review
@kybaker

Signed-off-by: Andrew Kofink <akofink@redhat.com>

### DIFF
--- a/src/AppConstants.js
+++ b/src/AppConstants.js
@@ -9,7 +9,8 @@ export const SID_SET = 'SID_SET';
 
 // Compliance App Constants
 export const COMPLIANCE_FETCH = 'COMPLIANCE_SUMMARY_FETCH';
-export const COMPLIANCE_FETCH_URL = `${BASE_URL}/compliance/profiles?search=has_test_results=true`;
+export const COMPLIANCE_SEARCH = encodeURIComponent('(has_policy_test_results=true AND external=false) OR (has_policy=false AND has_test_results=true)');
+export const COMPLIANCE_FETCH_URL = `${BASE_URL}/compliance/profiles?search=${COMPLIANCE_SEARCH}`;
 
 // Vulnerability App Constants
 const VULN_CVES = '/vulnerability/v1/report/executive';
--- a/src/SmartComponents/Compliance/ComplianceCard.js
+++ b/src/SmartComponents/Compliance/ComplianceCard.js
@@ -11,6 +11,7 @@ import {
 import React, { useEffect } from 'react';
 import { TemplateCard, TemplateCardBody, TemplateCardHeader } from '../../PresentationalComponents/Template/TemplateCard';
 import { chart_color_blue_200, chart_color_blue_300 } from '@patternfly/react-tokens';
+import { global_palette_black_300 } from '@patternfly/react-tokens/dist/esm/';
 import { supportsGlobalFilter, workloadsPropType } from '../../Utilities/Common';
 
 import { Button } from '@patternfly/react-core/dist/js/components/Button/Button';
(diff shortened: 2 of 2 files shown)


src/AppConstants.js
```diff
@@ -9,7 +9,8 @@ export const SID_SET = 'SID_SET';
 
 // Compliance App Constants
 export const COMPLIANCE_FETCH = 'COMPLIANCE_SUMMARY_FETCH';
-export const COMPLIANCE_FETCH_URL = `${BASE_URL}/compliance/profiles?search=has_test_results=true`;
+export const COMPLIANCE_SEARCH = '(has_policy_test_results = true AND external = false) OR (has_policy = false AND has_test_results = true)';
+export const COMPLIANCE_FETCH_URL = `${BASE_URL}/compliance/profiles?search=${COMPLIANCE_SEARCH}`;
```
vkrizan, line 13: Data passed through the URI params should be encoded. See [`encodeURI()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI).

AllenBW (approved):
Just the one thing already mentioned then okie dokie with it 🥰🦾

kybaker:
@akofink Can we show a gray circle (empty pie chart) if there is a 'zero state'? Like you see in the recommendations card.

akofink:
Updated with a hacky solution modeled after Advisor's for the empty PieChart.

src/SmartComponents/Compliance/ComplianceCard.js
```diff
@@ -78,14 +79,15 @@ const ComplianceCard = ({ fetchCompliance, complianceFetchStatus, complianceSumm
                                                             y: policy.attributes.compliant_host_count
                                                         }, {
                                                             x: 'Non-compliant',
-                                                            y: policy.attributes.total_host_count - policy.attributes.compliant_host_count
-                                                        }
+                                                            y: policy.attributes.test_result_host_count - policy.attributes.compliant_host_count
+                                                        }, (policy.attributes.test_result_host_count === 0 && { x: 'Compliant', y: '0' })
```
vkrizan, line 83: This seems to override previously defined "Compliant" category. Does this work because of the zero in a string?

src/AppConstants.js
akofink: Updated. Thanks!

src/SmartComponents/Compliance/ComplianceCard.js
akofink: This is the text shown when you hover over the chart. So when it's 0, it should show `Compliant: 0`. The y values aren't actually required to be unique.

kybaker:
Thanks for making the update! Looks good to me!

AllenBW (approved):
👍 🌮 💃

vkrizan (approved):

AllenBW:
off to prod-beta, will push to stable after that comes back 🅰️ 🆗

AllenBW:
@akofink this breaks the dashboard in beta, so its been reverted

https://github.com/RedHatInsights/insights-dashboard/pull/212/files#diff-d364033114a05cfb1ed690e122137a8a42d3919fee82c0246b2c07c6256eb8e8L115 appears to be the culprit, changing it to `)?.toFixed(1)` solves the breaking but not the borked ui

akofink:
@AllenBW the backend requirement for this PR isn't in prod yet. We plan to deploy it tomorrow, but it could slip to Thursday. Can we push this to QA for now so our QEs can verify it?

AllenBW:
AHHHH yah that woulda been good info to include in original pr @akofink can yah add the change mentioned (for the case where the var is undefined? ) n then we'll do a master merge which will include qa envs

akofink:
@AllenBW it's in the 'Additional Info' section of the description! :see_no_evil: No worries, though :laughing: Adding the safe navigation operator shouldn't be required IMO (I didn't change that line), but I can add it if you still think it's necessary.

AllenBW:
@akofink yah should check it out for yourself using current prod api see how it breaks, new pr

akofink:
Current prod API doesn't contain the changes required by this PR. It is expected to break there.
@AllenBW  I'm confused about how dashboard ships to different environments. I'd expect the same flow as all other services and frontends: CI -> QA -> prod. Why isn't it like that here? It sounds like you're shipping directly to prod beta and skipping CI and QA.

PR #223, opened 2021-01-29
Speed up build

Hyperkid123:
I've heard from @ryelo that the build takes a long time.

After checking it out locally, it took around 10 minutes for the production build to complete. I've managed to get it down to 50s.

The issue was that the build was bundling all pf react tokens and all react icons. That is around 5,5k modules which is the reason why the build was taking so long.

I've added transform imports plugins for the react icons and changed the imports if react tokens to use direct import paths. If you want you can add a transform imports plugin for react tokens as well. You can add it into babel.config.js file.

An important note here. Do not use shorthand imports for react-icons and tokens. Tree shaking does not work properly in these libraries.

### Before
screenshot-127 0 0 1_8888-2021 01 29-16_07_57

### After
screenshot-127 0 0 1_8888-2021 01 29-16_25_34


ryelo (approved):

Hyperkid123:
cc @mattnolting

Hyperkid123:
GOD DAMN YOU LINTER!

AllenBW (approved):
thats fast, yer awesome 🦞 🤗 🧼 :shipit:

PR #225, opened 2021-02-04
Use fec charts and translations packages to v3.

Hyperkid123:
@AllenBW is this app even using the frontend components charts? I see there is the gauge component referenced but it's not in the final build (at least I could not find it in the webpack build).

EDIT: I see it in the `GaugeSection` but that component is never used anyway.


AllenBW (approved):
Yup, yer right @Hyperkid123  in there never used, kinda ah thing  we kept around just in case ah dashboard card would need it

Hyperkid123:
@AllenBW I will first make a stable release of the translation package. I just wanted to check with you first.

Hyperkid123:
Ok its now a stable release

PR #226, opened 2021-02-08
Update deps  👩‍🏫🍔✂️

AllenBW:
Nathing breaks, all things updated


Hyperkid123:
@AllenBW please double check CSS in prod env :wink: we did some more CSS scoping before the sable release of the common webpack config.

PR #234, opened 2021-03-22
use correct link to vulnerability application

tkasparek:
image
This link should take user to the vulnerability landing page with some filters applied


AllenBW (approved):
🏅 💯 🙇 👍

PR #243, opened 2021-04-12
vulnerability: use correct fields for the data

tkasparek:
there's no need for fallback as all changes have been propagated to prod


AllenBW (approved):
beat me to it, @tkasparek is ah 📸 💘

PR #246, opened 2021-04-15
Respect beta path on Patch card

jiridostal:
Before: On beta, Patch links were pointing to stable
After: On beta, Patch links are linking to /beta


AllenBW (approved):
👍

PR #247, opened 2021-04-16
Respect beta path on Subscriptions card

cdcabrera:
## Prerequisites

## What
- Adds an `openshift` base constant
- Updates the Subscription card with relative path references for both `insights` plus `openshift` so they work with both `stable` and `beta` paths.

### Before
On beta, Subscription links were pointing to `stable`
### After
On beta, Subscription links are linking to `beta`

## Additional Info
Per discussion @rblackbu @bclarhk since Subscriptions doesn't have a landing page the default "app" link aims at RHEL

## Code Review

### DIFF
--- a/src/AppConstants.js
+++ b/src/AppConstants.js
@@ -2,6 +2,7 @@
 
 const BASE_URL = '/api';
 export const UI_BASE = './insights';
+export const UI_BASE_OPENSHIFT = './openshift';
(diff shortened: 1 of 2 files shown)


ryelo (approved):

AllenBW (approved):
LG2M 👍

mirekdlugosz:
Looks good on CI.
@AllenBW what do we need to do to have that on prod-beta, too?

PR #251, opened 2021-05-10
Update deps 🦑🤠❤️‍🔥

AllenBW:


AllenBW:
Buttttt this won't fix the micro tooltip we're seeing on charts

PR #256, opened 2021-06-18
vulnerability: update URL param

tkasparek:
there was a filtering in update on Vulnerability side - Dashboard needs
to be switched as well. This is 1st commit which is backwards compatible
and also ensures nothing will break once Vulnerability updates
production.
This commit will be followed by another one once Vulnerability
production is updated which will remove the part used for compatibility.


AllenBW (approved):
true,true kinda seems odd to me, but yah know what yer doing @tkasparek ! This good for prod stable now?

tkasparek:
> true,true kinda seems odd to me, but yah know what yer doing @tkasparek ! This good for prod stable now?

That's the backward compatible solution, bit nasty, but nothing explodes when user clicks on the link and actually gets into vulnerability.

tkasparek:
@AllenBW we'll update our prod probably next, so it's completely up to you if you want to wait for us, or release it now. Nothing should break in any case.

PR #258, opened 2021-07-15
Update URL parameters on vulnerability card

leSamo:
## Prerequisites

## What
Follow-up to #256. Update `affecting` URL parameter on vulnerability card. Vulnerability backend has been updated (all envs including prod) and there no need for this backwards compatibility. Fixes this weirdness:
1ormore

## Code Review


AllenBW (approved):
LG2M! Off we go 🧜‍♀️

PR #262, opened 2021-09-10
Hide subscriptions card in fedramp env.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-16048

This is part of a FedRamp effort. More details can be found here: https://issues.redhat.com/browse/RHCLOUD-14323

### DIFF
--- a/src/PresentationalComponents/Dashboard/Dashboard.js
+++ b/src/PresentationalComponents/Dashboard/Dashboard.js
@@ -3,6 +3,7 @@ import './dashboard.scss';
 import { Grid, GridItem } from '@patternfly/react-core/dist/esm/layouts';
 import { PageSection, PageSectionVariants, Title } from '@patternfly/react-core/dist/esm/components';
 import React, { Suspense, lazy, useContext, useEffect, useState } from 'react';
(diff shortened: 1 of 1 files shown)


src/PresentationalComponents/Dashboard/Dashboard.js
```diff
@@ -100,7 +102,7 @@ const Dashboard = ({ workloads }) => {
                                 }
                             </Suspense>
                             <Suspense fallback={ <Loading /> }>
-                                {permission.subscriptions &&
+                                {!isFedramp && permission.subscriptions &&
```
AllenBW, line 105: Can we just kill the whole suspense component? seems kinda odd to load for ah thing we wont be showing

AllenBW: <3

AllenBW (approved):
Noice, that'll hide it <3 🙇

AllenBW:
@Hyperkid123 this work to be pushed all the way to prod-stable?

Hyperkid123:
Yeah, It should. The fedramp flag is not in chrome prod yet, but that's OK. If it's missing, it will just render the tile. It will start working eventually.

PR #274, opened 2021-12-22
Use webpack proxy at all times and remove html webpack plugin

karelhala:
### Add `HttpsProxyAgent`

Since we want to proxy to different API (for instance when running API locally) let's add `'http://squid.corp.redhat.com:3128'` to it so we won't fail when querying entitlements API.

### Use proxy at all times

Since webpack proxy has matured and should be used almost exclusively let's use it. Also update env variables to switch between environments and beta/non-beta.

### Filter out HTML webpack plugin

We don't need this plugin anymore, chrome should be serving single HTML from now on.

### Remove unsed webpack and webpack-cli deps

OUr shared config should install them all the times.


rvsia (approved):

config/dev.webpack.config.js
```diff
@@ -2,13 +2,14 @@
 const { resolve } = require('path');
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 const config = require('@redhat-cloud-services/frontend-components-config');
+const HttpsProxyAgent = require('https-proxy-agent');
```
Hyperkid123, line 5: This proxy agent is a part of the common config. Are you sure we need it here again?

config/dev.webpack.config.js
```diff
@@ -28,8 +30,10 @@ plugins.push(
     })
 );
 
+webpackConfig.devServer.client.overlay = false;
+
 module.exports = (env) => {
     env && env.analyze === 'true' && plugins.push(new BundleAnalyzerPlugin());
 
-    return { ...webpackConfig, plugins };
+    return { ...webpackConfig, plugins: plugins.filter((plugin) => plugin.constructor.name !== 'HtmlWebpackPlugin') };
```
Hyperkid123, line 38: HTML plugin was also changed in the latest version. The JS injection was disabled.

config/dev.webpack.config.js
karelhala: Yup, because there's a env flag `API_ENDOINT` which deals with redirecting to custom API, however that requires `squid` proxy for prod and stage envs.

config/dev.webpack.config.js
karelhala: The problem here was with ESI, for some reason when running locally it just broke and wouldn't show any content. Once the plugin was disabled it started to run.

Hyperkid123 (approved):

PR #280, opened 2022-01-13
Fix empty screen by locking version of FEC

karelhala:


Hyperkid123:
This will not fix the issue. I will open PR with the new version and local chrome template to fix it.

PR #281, opened 2022-01-13
Use chrome HTML to fix failing ESI tag mutations.

Hyperkid123:
There are ongoing issues caused by either `webpack-dev-server` itself or one of its dependencies with mutating the ESI tags inside the HTML templates with the actual content.

We do not know the cause of the issue nor how to fix it. This PR uses a chrome HTML template directly in local env. As this is approach will be used on products in the future, we can fast track it in this repository to unblock the development.

### DIFF
--- a/config/dev.webpack.config.js
+++ b/config/dev.webpack.config.js
@@ -2,7 +2,6 @@
 const { resolve } = require('path');
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 const config = require('@redhat-cloud-services/frontend-components-config');
-const HttpsProxyAgent = require('https-proxy-agent');
 const { config: webpackConfig, plugins } = config({
     rootFolder: resolve(__dirname, '../'),
     debug: true,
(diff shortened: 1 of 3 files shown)


config/dev.webpack.config.js
```diff
@@ -19,7 +20,6 @@ const { config: webpackConfig, plugins } = config({
             changeOrigin: true,
             autoRewrite: true,
             ws: true,
-            ...['prod', 'stage'].includes(process.env.ENVIRONMENT) && { agent: new HttpsProxyAgent('http://squid.corp.redhat.com:3128') }
```
karelhala, line 22: Isn't this still required when user runs custom local API?

Hyperkid123: Nope. That is already a part of the webpack proxy

Hyperkid123: https://github.com/RedHatInsights/frontend-components/blob/master/packages/config-utils/proxy.js#L58

karelhala: Yeah, I know it's part of the general proxy. But previously there were errors when this config was missing from the customProxy as well. Is there any benefit of removing it?

Hyperkid123: I don't have the local API setup. I would need someone who has to test it. I think having the proxy agent active twice might have some unwanted issues.

karelhala: Sure thing, let's remove it. If someone points to it that it won't work we'll re-add it.

karelhala (approved):

PR #285, opened 2022-01-19
RHIROS-488 updated ROS card to add individual links

PreetiW:
### Description:

Updated the Resource Optimization (ROS) card in the dashboard with numbered links. 

I have created the [ROS side PR](https://github.com/RedHatInsights/ros-frontend/pull/89) here. If everything looks good we can merge this as until [ROS PR](https://github.com/RedHatInsights/ros-frontend/pull/89) is not merged it wont work end to end but not processing query params won't break anything.

**jira link:** https://issues.redhat.com/browse/RHIROS-447

### Video:
https://drive.google.com/file/d/1-OCczzQYK_lDo2sSyO7I3bbGfEf_UNZf/view?usp=sharing

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,10 @@
   "name": "insights-dashboard",
   "version": "1.0.0",
   "private": false,
+  "engines": {
+    "node": ">=16.0.0",
+    "npm": ">=7.0.0"
+  },
   "repository": {
     "type": "git",
     "url": "https://github.com/RedHatInsights/insights-dashboard"
--- a/src/SmartComponents/ResourceOptimization/ResourceOptimizationCard.js
+++ b/src/SmartComponents/ResourceOptimization/ResourceOptimizationCard.js
@@ -25,8 +25,8 @@ const ResourceOptimizationCard = ({
     }, [fetchRosIsConfigured]);
 
     const navigateTo = useChromePush();
-    const waitingForDataUrl = `${UI_BASE}/ros?state=waiting`;
-    const suggestionsUrl = `${UI_BASE}/ros?state=suggestions`;
+    const waitingForDataUrl = `${UI_BASE}/ros?with_waiting_for_data=true`;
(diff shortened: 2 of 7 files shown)


src/SmartComponents/ResourceOptimization/ResourceOptimizationCard.js
```diff
@@ -25,6 +25,9 @@ const ResourceOptimizationCard = ({
     }, [fetchRosIsConfigured]);
 
     const navigateTo = useChromePush();
+    const waitingForDataUrl = `${UI_BASE}/ros?state=waiting`;
+    const suggestionsUrl = `${UI_BASE}/ros?state=suggestions`;
```
kgaikwad, line 29:
```suggestion
    const waitingForDataUrl = `${UI_BASE}/ros?with_waiting_for_data=true`;
    const suggestionsUrl = `${UI_BASE}/ros?with_suggestions=true`;
```

Instead of `state` key in query param, will it be good idea to use query string flags here as those are not state values?
Example - `${UI_BASE}/ros?with_suggestions=true`;

Any thoughts?

PreetiW: I agree @kgaikwad it will make it more readable, I will update it and push the changes soon.

PreetiW (commented):
@kgaikwad I have updated the PR as per your suggestion. Please re-review it.

kgaikwad (approved):
This looks good to me!

package.json
```diff
@@ -2,6 +2,10 @@
   "name": "insights-dashboard",
   "version": "1.0.0",
   "private": false,
+  "engines": {
+    "node": ">=16.0.0",
```
Hyperkid123, line 6: If not already you should also update the `travis.yml` to run on 16 so your production build uses the same version.

PreetiW: Good point @Hyperkid123 💯 have pushed the change in the latest commit.

Hyperkid123 (approved):

PR #286, opened 2022-01-19
fix: VULN-2121:vulnerability_results:read permission should be suffic…

MichalSajdik:
…ient for vulnerability dashboard

Jira: https://issues.redhat.com/browse/VULN-2121


leSamo (approved):
This is a good temporary solution, but I think the permissions systems should be reworked to match * in permission with anything.

PR #290, opened 2022-02-01
[DRFT-1006] Add systems to compare button

jpramos123:
Previously we were adding only the baseline to the compare table, this PR will add associated system ids to the compare table as well.

We are limiting up to 1 baseline and 4 systems to be added to the drift table.

Added a tooltip to the button as well

Thanks


bastilian (dismissed):
LGTM! Thank you @jpramos123!

bastilian (approved):

PR #292, opened 2022-02-01
[DRFT-1014] Adjust drift card title and hide dropdown when card is not expanded

jpramos123:


bastilian (approved):
Thank you @jpramos123!

PR #302, opened 2022-03-24
Add AAP and MSSQL global filter support

johnsonm325:
This PR adds support for the Ansible Automation Platform and Microsoft SQL global filters.

I have removed the empty state for the time being as we will be refactoring some of the code in Dashboard to improve global filters.

### DIFF
--- a/src/PresentationalComponents/Dashboard/Dashboard.js
+++ b/src/PresentationalComponents/Dashboard/Dashboard.js
@@ -2,15 +2,12 @@ import './dashboard.scss';
 
 import { Grid, GridItem } from '@patternfly/react-core/dist/esm/layouts';
 import { PageSection, PageSectionVariants, Title } from '@patternfly/react-core/dist/esm/components';
-import React, { Suspense, lazy, useContext, useEffect, useState } from 'react';
+import React, { Suspense, lazy, useContext } from 'react';
 import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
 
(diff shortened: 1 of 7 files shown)


johnsonm325:
@bastilian @leSamo Compliance and Vulnerability cards do no like the global filters in dashboard. It seems like maybe there's something wrong with the fetches? Maybe this is the intent?

Vulnerability: https://github.com/RedHatInsights/insights-dashboard/blob/master/src/SmartComponents/Vulnerability/VulnerabilityCard.js#L76
- I tried to add the filters for AAP and MSSQL here, but that seemed to break it. https://github.com/RedHatInsights/insights-dashboard/blob/master/src/SmartComponents/Vulnerability/VulnerabilityCard.js#L71-L75


Compliance: https://github.com/RedHatInsights/insights-dashboard/blob/master/src/SmartComponents/Compliance/ComplianceCard.js#L70
- If I chose a global filter, this one renders with the following message
image
- This seems to happen in prod as well, so I think it's an existing issue that I'm not sure if it's intentional.

leSamo:
@johnsonm325 Vulnerability does not yet support AAP and MSSQL workloads on the backend, but it's coming very soon hopefully

bastilian:
@johnsonm325 Compliance does not support global filters at all.

MichalSajdik (approved):
LGTM, we need to fix this at least for dashboard for now, because of https://issues.redhat.com/browse/ESSNTL-2426

PR #308, opened 2022-04-20
Add release scripts

johnsonm325:
These release scripts may make releasing more fail-safe. Let me know if I've missed something.


mkholjuraev (commented):
I really like this. Regarding the code, How about having one script file for all environments, then we can pass in environment name into the script and make a release for that env. I am not sure if this can be done. This is just an idea.

leSamo:
Closing this as it's become obsolete with containerized builds.

PR #315, opened 2022-06-27
Update packages for working snapshot tests

AsToNlele:
Had to update packages like in other apps. Also added packages which will be needed for testing.

### DIFF
--- a/config/setupTests.js
+++ b/config/setupTests.js
@@ -1,5 +1,5 @@
 import { configure, mount, render, shallow } from 'enzyme';
-import Adapter from 'enzyme-adapter-react-16';
+import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
 import React from 'react';
 
 configure({ adapter: new Adapter() });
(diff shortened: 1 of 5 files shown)


AsToNlele:
after this is merged I will push a test for AdvisorCard

gabipodolnikova (changes requested):
I found just one small styling issue on the Drift card - under `Investigate in Drift` button there is no longer a padding.

Also could you please fix the console errors? It should be enough to remove having `key` as an index. Just search where `DataListCell` is used and replace it with something that has more description to it :D 

Otherwise it looks great, it will also help to fix a styling issue in Inventory with this Frontend components update, so thank you!

AsToNlele:
Thanks for your review! I've fixed the console errors, but I don't see the styling issue on the Drift card 😿

image

gabipodolnikova (approved):
I don't see the styling issue now either, weird :D So great, LGTM! :)

PR #316, opened 2022-07-01
Add Advisor test

AsToNlele:
[ADVISOR-2058](https://issues.redhat.com/browse/ADVISOR-2058)

Added a snapshot test for the advisor card


mkholjuraev (commented):
I am wondering if the snapshot testing is sufficient or if a few unit tests are needed to cover the button, loading the component. I see that the component is tiny though.

PR #325, opened 2022-11-10
feat(codecov):fixing codecoverage

Fewwy:
The codecov.io reports were missing it's content.
I checked the settings and I think I fixed it


package.json
```diff
@@ -27,7 +27,7 @@
     "start:proxy:beta": "NODE_ENV=development BETA=true webpack serve --config config/dev.webpack.config.js",
     "start": "npm run start:proxy",
     "start:beta": "npm run start:proxy:beta",
-    "test": "jest --verbose",
+    "test": "jest --coverage --verbose",
```
bastilian, line 30: Shouldn't `"collectCoverage": true,` take care of this without `--coverage`

gkarat: I was about to write the same thing. `jest` entry in package.json should already modify the behavior of jest.

Fewwy: Thanks! Updated the PR

gkarat (approved):

PR #328, opened 2022-11-21
feat(ZeroState): ESSNTL-3647 - add dedicated static page for the zero state content

tompsota:
resolves [ESSNTL-3647](https://issues.redhat.com/browse/ESSNTL-3647)


bastilian (dismissed):
LGTM! Thank you @tompsota!

src/AppEntry.js
```diff
@@ -4,14 +4,13 @@ import IntlProvider from '@redhat-cloud-services/frontend-components-translation
...
 import { BrowserRouter as Router } from 'react-router-dom';
-import { getBaseName } from '@redhat-cloud-services/frontend-components-utilities/helpers/helpers';
 import { init } from './Store';
 import messages from '../locales/data.json';
 
 const AppEntry = () =>
     <IntlProvider locale={ navigator.language.slice(0, 2) } messages={ messages } onError={ console.log }>
         <Provider store={ init().getStore() }>
-            <Router basename={ getBaseName(window.location.pathname) }>
+            <Router basename={ '/insights' }>
```
gkarat, line 13: Running your PR, I wasn't able to open /start in beta. Consider using this:
```suggestion
            <Router basename={ getBaseName(window.location.pathname, 1) }>
```
ref: https://github.com/RedHatInsights/frontend-components/blob/3c0e6017829b98a78c92e2c2063d114c45013933/packages/utils/src/helpers/helpers.ts#L37

gkarat (changes requested):

src/AppEntry.js
bastilian: @gkarat It is available at `/insights/start` which, looking through the Jira(s), seems to be the intention.

johnsonm325 (approved):
This looks good. I assume that we're simply creating a url that links to the page at this point and at a later date determining a way to direct the user there when they have no systems set up? Or a way to determine a first-time user?

src/AppEntry.js
gkarat: yeah, you are true. however, it often turns out PMs also wanted it for beta, and just to avoid future jiras/complaints, we can catch two rabbits now :-)

gkarat (approved):

PR #332, opened 2022-12-13
fix(ESSNTL-3956): remove stale and fresh params from inventory url

mkholjuraev:
Resolves: [ESSNTL-3956](https://issues.redhat.com/browse/ESSNTL-3956);

Description:
The number of systems is different in the Inventory applcation when navigated from Dashboard 'Systems registered with Insights' link.

To test:

1. Visit dashboard
2. Observe the number of systems that has insights.
3. Click on the link 'Systems registered with Insights' that will bring you to inventory app.
4. Observe the number of systems on the table is the same.


leSamo (approved):
LGTM! The numbers do match up.

PR #338, opened 2023-01-10
[ESSNTL-3976] Dashboard links account for beta

johnsonm325:
steps to reproduce:
1. visit https://console.redhat.com/beta/insights/dashboard on beta version
2. click on any app link. Example: "Systems registered with Insights"

You'll notice that the `/beta` in the url is dropped. With this fix, the beta remains.

For this PR, check and make sure that both beta and stable redirect to the proper url with link clicks.


src/Utilities/hooks/useChromePush.js
```diff
@@ -5,7 +5,11 @@ export const useChromePush = () => {
     return (e, link) => {
         if (typeof push === 'function') {
             e.preventDefault();
-            push(link.replace(/^.\//, '/'));
+            if (insights.chrome.isBeta()) {
```
mkholjuraev, line 8: I think we can remove the returned function as it is not needed. 

What this does now is whenever the dashboard unmounts, it either adds or removes `/beta`. I think this is not needed if we want to keep the environment the same across the navigation.

mkholjuraev (approved):

johnsonm325:
@mkholjuraev might want to take another look at this. It appears that the `useChromePush` wasn't needed in the first place. I removed all instances of it. Take a look and let me know if I broke something :)

mkholjuraev (approved):
It seems you are right, but I think we need to get this PR verified by QA thoroughly.

johnsonm325:
> It seems you are right, but I think we need to get this PR verified by QA thoroughly.

Do you think we need to merge first? Or have QA look at it as a PR?

mkholjuraev:
> > It seems you are right, but I think we need to get this PR verified by QA thoroughly.
> 
> Do you think we need to merge first? Or have QA look at it as a PR?

Aha, get it verified after merging.

PR #344, opened 2023-01-23
Reduce calls to Patch advisories to a single request

PaulWay:
The subtotals for the three advisory types - bugs, security and enhancements - are available in the `meta` structure of the request. This means that we do not have to make three requests to the Patch API in order to get the data necessary to build the Advisory Types pie chart.

This work basically generates a single request and stores the subtotals data.  The subtotal for each advisory type are then pulled out in the PatchManagerCard component.

Signed-off-by: Paul Wayper <paulway@redhat.com>

### DIFF
--- a/src/SmartComponents/PatchManager/PatchManagerCard.js
+++ b/src/SmartComponents/PatchManager/PatchManagerCard.js
@@ -31,7 +31,7 @@ const PatchManagerCard = ({
 }) => {
     const intl = useIntl();
     const isLoaded = [systemsStatus, advisoriesStatus].every(item => item === 'fulfilled');
-    const security = advisories.security;
+    const { security, bugfix: bugs, enhancement: enhancements } = advisories || {};
     const bugs = advisories.bugfix;
     const enhancements = advisories.enhancement;
     const pieChartData = [
--- a/src/SmartComponents/PatchManager/PatchManagerCard.js
+++ b/src/SmartComponents/PatchManager/PatchManagerCard.js
@@ -5,7 +5,7 @@ import { PATCHMAN_ID, UI_BASE } from '../../AppConstants';
 import React, { useEffect } from 'react';
 import { TemplateCard, TemplateCardBody, TemplateCardHeader } from '../../PresentationalComponents/Template/TemplateCard';
 import { capitalize, globalFilters, workloadsPropType } from '../../Utilities/Common';
-import { patchmanFetchBugs, patchmanFetchEnhancements, patchmanFetchSecurity, patchmanFetchSystems } from '../../AppActions';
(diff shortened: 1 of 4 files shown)


src/SmartComponents/PatchManager/PatchManagerCard.js
```diff
@@ -22,13 +22,18 @@ import messages from '../../Messages';
...
-    selectedTags, workloads, SID }) => {
+const PatchManagerCard = ({
+    systems, systemsStatus, fetchSystems,
+    advisories, advisoriesStatus, fetchAdvisories,
+    selectedTags, workloads, SID
+}) => {
     const intl = useIntl();
-    const isLoaded = [systemsStatus, securityStatus, bugsStatus, enhancementsStatus].every(item => item === 'fulfilled');
+    const isLoaded = [systemsStatus, advisoriesStatus].every(item => item === 'fulfilled');
+    const security = advisories.security;
```
mkholjuraev, line 34: As @gkarat mentioned, we need to handle the initial render when `advisories` is undefined. We can achieve this with object destruction: 
```suggestion
    const { security, bugfix: bugs, enhancement: enhancements } = advisories || {};
```

src/SmartComponents/PatchManager/PatchManagerCard.js
```diff
@@ -22,13 +22,18 @@ import messages from '../../Messages';
...
+    const bugs = advisories.bugfix;
+    const enhancements = advisories.enhancement;
```
gkarat, line 36: `advisories` are undefined before the network request is fulfilled. This leads to an error that breaks the whole page component because you are accessing `security` and further parameters on undefined type.

The very first thing came to mind and that fixes this is to provide an alternative value: 
```suggestion
    const security = advisories?.security || 0;
    const bugs = advisories?.bugfix || 0;
    const enhancements = advisories?.enhancement || 0;
```

gkarat (changes requested):
Hi! Thanks for your PR! when deployed locally with this PR, the page crashed. but I found the cause of the problem and described it below.

but overall LGTM, it's a good improvement

mkholjuraev (approved):
LGTM! Works as expected. @PaulWay Thank you.

mkholjuraev (approved):

PaulWay:
How is this looking - does anything else need to be done before it can be merged?

src/SmartComponents/PatchManager/PatchManagerCard.js
PaulWay: I think this is fixed with `const { security, bugfix: bugs, enhancement: enhancements } = advisories || {};` now?

mkholjuraev:
I have approved this PR, but there is one pending change request by @gkarat .

PR #346, opened 2023-02-28
Remove moment.js library

mhuth:
The moment.js library is rather large and some network bandwidth can be saved by removing it from the project.  Its use is only minimal and its use replaced by functions from the built in Date library.


mkholjuraev (approved):
LGTM!

PR #351, opened 2023-03-15
fix: Fix the non-working Drift link button

gkarat:
Before, the button didn't behave as a link and didn't do anything after click.

image


bastilian (approved):
LGTM! Thank you for the fix @gkarat!

PR #398, opened 2023-05-15
Remove pie chart from latest critical notifications

leSamo:
Ticket: https://issues.redhat.com/browse/VULN-2576
Mockup: https://www.sketch.com/s/4d721f7e-f57e-46e6-ba1c-739c0cad040e/a/09e9lyb?annotation=8acba49e-e913-4872-8fa5-f9b9287bc420#Comment

- removed the "View more" button because it navigated user to the detail of first listed CVE which was confusing for the users when there were multiple CVEs listed
- replaced pie chart with just a number of systems exposed
- added a text explaining how the number of systems exposed is calculated

## Before:
Screenshot from 2023-05-15 13-18-24

## After:
Screenshot from 2023-05-15 13-09-31

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -707,7 +707,12 @@ export default defineMessages({
     systemsExposed: {
         id: 'systemsExposed',
         description: 'Systems exposed',
-        defaultMessage: 'Systems exposed'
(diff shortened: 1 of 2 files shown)


AsToNlele (approved):
LGTM, thank you :smile:

PR #404, opened 2023-06-06
fix(ComplianceDoc): RHICOMPL-3781 - Fix unsupported doc link

LightOfHeaven1994:
Fixing non-working link to point on Supported SSGs article on Zero state
image (14)


adonispuente (approved):
LGTM!

PR #405, opened 2023-06-10
feat(cypress): command to mount components easily

Fewwy:
I added a cypress command that allows to mount component easier and provide components with suspense

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -24,24 +24,19 @@
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
 // one of the fec dependencies talks to window.insights.chrome
-import React, { Suspense } from 'react';
+import React from 'react';
 import { Provider } from 'react-redux';
 import { init } from '../../src/Store';
 import messages from '../../locales/data.json';
 import IntlProvider from '@redhat-cloud-services/frontend-components-translations/Provider';
 import { mount } from '@cypress/react';
 
-Cypress.Commands.add('mountWithContext', (Component, FallBackComponent) => {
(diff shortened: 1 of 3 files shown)


cypress/support/commands.js
```diff
@@ -24,6 +24,29 @@
...
+
+Cypress.Commands.add('mountWithContext', (Component, FallBackComponent) => {
+
+    return mount(
+        <IntlProvider messages={messages}>
+            <Provider store={ init().getStore() }>
+                {FallBackComponent ?
+                    (<Suspense fallback={FallBackComponent}>
+                        <Component />
+                    </Suspense>) :
```
gkarat, line 42: What's the reason for explicit Suspense support here? You could use the Component argument and pass anything you want inside of it including components wrapped in Suspense. This would help make the function more generic...

cypress/support/commands.js
```diff
@@ -24,6 +24,29 @@
...
 // Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
 // one of the fec dependencies talks to window.insights.chrome
+import React, { Suspense } from 'react';
+import { Provider } from 'react-redux';
+import { init } from '../../src/Store';
+import messages from '../../locales/data.json';
+import IntlProvider from '@redhat-cloud-services/frontend-components-translations/Provider';
+import { mount } from '@cypress/react';
+
+Cypress.Commands.add('mountWithContext', (Component, FallBackComponent) => {
```
gkarat, line 34: Can we also support passing props for Component? Like here for example https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/cypress/support/commands.js#L35

gkarat (commented):
looks good, but would be nice to add some improvements

gkarat (approved):

PR #406, opened 2023-06-10
Preparing workshop

Fewwy:


src/SmartComponents/Vulnerability/VulnerabilityCard.cy.js
```diff
@@ -0,0 +1,93 @@
...
+    beforeEach(() => {
+        cvesInterceptors['failed with server error']();
+        cy.mountWithContext(
+            VulnerabilityCard,
+            Loading,
+            Masonry,
+            { breakpointCols: breakpointColumnsObj,
+                className: 'ins-l-masonry',
+                columnClassName: 'ins-l-masonry_column' },
+            true);});
```
gkarat, line 85: mountWithContext only wraps the required component in all required routers/providers, etc. it shouldn't care about children nor handle some logic that children can do by themselves.

gkarat (changes requested):

PR #417, opened 2023-06-28
fix(RHINENG-734): Application zero states are offering a membership

jiridostal:


bastilian (approved):
LGTM! Thank you @jiridostal!

PR #445, opened 2023-09-27
Cypress tests and interceptors

Fewwy:
Added new interceptors and tests


src/SmartComponents/Advisor/Advisor.cy.js
```diff
@@ -0,0 +1,59 @@
...
+        { name: 'Availability', value: '62' },
+        { name: 'Stability', value: '10' },
+        { name: 'Performance', value: '20' },
+        { name: 'Security', value: '35' }]
+};
+
+const AdvisorCard = () => {
+    return (
+        <Grid hasGutter>
+            <Masonry
```
bastilian, line 28: What is this for?

Fewwy: I am replicating the component structure in the Dashboard.js
image

bastilian: It might be better to just wrap the `VulnerabilityCard` only in what it really needs, likely just the `Router` and maybe a store provider.

Introducing more than needed, means more to investigate if something fails.

Fewwy: We can remove Grid component but we have to keep Masonry, there are at least 2 reasons for that.
1) If it breaks on the Grid or Masonry level we will know about that
2) It looks different from the actual representation without the Masonry component. 
This is how it looks with Masonry
image

And this is without it
image

bastilian: @Fewwy Other than the width, I see no visual difference. and it just being wider is very acceptable.

Fewwy: Ok :) I will remove both Grid and Masonry components then

Fewwy: @bastilian I removed the masonry and grid from this and 2 other files

Fewwy:
@gkarat It is merged in the other PR. We can close this one

PR #446, opened 2023-09-29
fix(ESSNTL-3831): Links without page refresh

AsToNlele:
[ESSNTL-3831](https://issues.redhat.com/browse/ESSNTL-3831)

Links now work without full page refresh, so eg. global filter should stay applied.

How to test:
- Make sure all links to our apps are dynamic


bastilian (approved):
NICE! Thank you @AsToNlele!

src/PresentationalComponents/ZeroState/AppZeroState.js
```diff
@@ -19,6 +19,7 @@ const AppZeroState = ({ app, customInstructions, customButton, customText, custo
                     customTitle={customTitle}
                     appId={appId}
                 />
+                xd
```
mkholjuraev, line 22: Before this gets merged, let's remove this.

mkholjuraev (changes requested):
It seems there is a small leftover :). For visibility, I am requesting changes

src/PresentationalComponents/ZeroState/AppZeroState.js
bastilian: Nice catch. :D

AsToNlele: Thanks this one debugging tool got away! :smile:

AsToNlele:
Removed the `xd`, thanks for reviews!

PR #455, opened 2023-12-11
fix(RHINENG-5981): vulnerability should use new hybrid affected filte…

mkholjuraev:
…r values conditionally

Resolves: https://issues.redhat.com/browse/RHINENG-5981

The links in Vulnerability section of dashboard to view security rules and known exploits need to have affecting=rpmdnf%2Cedge in them when edge parity feature flag is enabled and there is an edge devices in the account level, otherwise the redirect will have an error as these are required from API now


xmicha82 (approved):
LGTM, thx @mkholjuraev

PR #456, opened 2023-12-13
fix(SystemInventoryHeader): RHINENG-2743 total include edge systems

xmicha82:


src/AppActions.js
```diff
@@ -61,10 +61,21 @@ export const fetchInventorySummary = (options) => ({
     payload: fetchData(ActionTypes.INVENTORY_FETCH_URL, {}, options)
 });
 
-export const fetchInventoryTotalSummary = (options) => ({
-    type: ActionTypes.INVENTORY_TOTAL_FETCH,
-    payload: fetchData(ActionTypes.INVENTORY_TOTAL_FETCH_URL, {}, options)
-});
+export const fetchInventoryTotalSummary = (options, edgeFeatureFlag) => {
```
mkholjuraev, line 64: It looks like this function is never called. While testing locally, as it is not called, I did not observe this new API request being made. As a result, the total number is still the old conventional systems' total number

mkholjuraev (approved):
LGTM! Verified both flag states and the API request. The number is consistent with the expectations. @xmicha82 thank you!

PR #457, opened 2024-01-18
fix(InventoryHeader): RHINENG-2743 fix displayed totals

xmicha82:
Fixed totals to match sum of systems in inventory

### DIFF
--- a/cypress/fixtures/edgeHosts.json
+++ b/cypress/fixtures/edgeHosts.json
@@ -1,0 +1,6 @@
+{
+  "results": [],
+  "count": 250,
+  "limit": 1,
+  "offset": 0
+}
--- a/cypress/fixtures/featureFlags.json
+++ b/cypress/fixtures/featureFlags.json
@@ -1,0 +1,12 @@
+{
+  "toggles": [
+      {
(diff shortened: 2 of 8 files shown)


bastilian:
@xmicha82 Would you mind adding a test for this? There is already cypress setup for the header. The test just needs to verify that the correct requests are called.

cypress/fixtures/edgeHosts.json
```diff
@@ -0,0 +1,6 @@
+{
+  "results": [],
```
mkholjuraev, line 2: Should we add some edge hosts here?

mkholjuraev (approved):
LGTM

xmicha82:
@mkholjuraev the thing is, the tests behave non-deterministically, sometimes they fail, sometimes they work, without any change 🤔 
and they seem to fail on network requests, for some reason

cypress/fixtures/edgeHosts.json
xmicha82: i mean, we don't need it for anything right now, it could be added in case it's needed for the tests
but i can add a few of them if you think it's better that way

PR #465, opened 2024-02-15
fix(InventoryHeader): RHINENG-2743 replace edge totals with edge warning

xmicha82:
Instead of inventory totals containing edge devices, display warning that the devices are not included

### DIFF
--- a/src/SmartComponents/SystemInventory/EdgeDevicesWarning.js
+++ b/src/SmartComponents/SystemInventory/EdgeDevicesWarning.js
@@ -1,0 +1,31 @@
+import React, { useEffect } from 'react';
+import { useIntl } from 'react-intl';
+import { Alert } from '@patternfly/react-core';
+import messages from '../../Messages';
+import { useFeatureFlag } from '../../Utilities/Hooks';
+import { useDispatch, useSelector } from 'react-redux';
+import * as AppActions from '../../AppActions';
+
+export const EdgeDevicesWarning = () => {
+    const intl = useIntl();
+    const isEdgeParityEnabled = useFeatureFlag('edgeParity.inventory-list');
+    const { edgeTotalSummary, edgeTotalFetchStatus } = useSelector(({ DashboardStore }) => ({
+        edgeTotalSummary: DashboardStore.edgeTotalSummary,
+        edgeTotalFetchStatus: DashboardStore.edgeTotalFetchStatus
+    }));
+    const dispatch = useDispatch();
+
+    useEffect(() => {
+        dispatch(AppActions.fetchEdgeTotal({}));
+    }, [dispatch]);
+
+    return isEdgeParityEnabled &&
+        edgeTotalFetchStatus === 'fulfilled' &&
+        edgeTotalSummary.count > 0
+        ? <Alert
+            variant="info"
+            isInline
+            title={intl.formatMessage(messages.edgeWarning)}
+        />
+        : null;
+};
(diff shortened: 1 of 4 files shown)


src/SmartComponents/SystemInventory/EdgeDevicesWarning.js
```diff
@@ -0,0 +1,43 @@
...
+        : null;
+};
+
+EdgeDevicesWarning.propTypes = {
+    fetchEdgeTotal: PropTypes.func,
+    edgeTotalSummary: PropTypes.object,
+    edgeTotalFetchStatus: PropTypes.string
+};
+
+export default connect(
```
bastilian, line 35: Prefer hooks like `useDispatch` and `useSelector` instead.

src/SmartComponents/SystemInventory/SystemInventoryHeader.js
```diff
@@ -75,92 +70,95 @@ const SystemInventoryHeader = ({
...
-                                <Button
-                                    variant='primary'
-                                >
-                                    { intl.formatMessage(messages.systemInventoryCTA) }
-                                </Button>
-                            </InsightsLink>
-                        </FlexItem>
                     </Flex>
-                </Flex>
+                    <EdgeDevicesWarning />
```
bastilian, line 160: Looking at the Vulnerability Systems page, which is taken as an example for this message, I think the warning should rather be in the content area and not the header part of the page.

xmicha82: according to the discussion on the ticket: https://issues.redhat.com/browse/RHINENG-2743 the UX placed the banner inside, because it only affects the header content, not the whole page
it does not look the best, that is true, but i don't see any other solution which would communicate the same thing

LightOfHeaven1994: @xmicha82 My 5 cents regarding alert placement, According to the mock alert should be placed above inventory data 

Currently: 
image

Expected:
image

xmicha82: yeah, i know about this, i was just waiting for the resolution from Andy Burka on where to place it

LightOfHeaven1994: @xmicha82 Last piece is the message itself - should be `"Immutable (OSTree) systems are included in the inventory totals."` instead of `"Immutable (OSTree) systems data is not included in the dashboard."`

src/SmartComponents/SystemInventory/SystemInventoryHeader.js
```diff
@@ -75,92 +70,95 @@ const SystemInventoryHeader = ({
...
-                                            { count: inventoryWarningSummary.total || 0 }
-                                        ) }
-                                        state="critical"
-                                        systemInventory
-                                    />
-                                </InsightsLink>
+                                    link='/?source=puptoo'
+                                />
+                            }
+                            {/* {inventoryFetchStatus === 'fulfilled' && inventoryTotalFetchStatus === 'fulfilled' &&
```
mkholjuraev, line 91: I believe we can remove this commented out code

mkholjuraev (approved):
LGTM! @xmicha82 thank you. Let's remove those code that is commented out before merging.

mkholjuraev:
@xmicha82 merging this to get QA verification and finally close out the edge parity epic

PR #472, opened 2024-03-07
feat(RHINENG-8744): show only conventional systems count in inventory…

mkholjuraev:
Resolves: https://issues.redhat.com/browse/RHINENG-8744

To test:
1. visit dashboard landing page
2. observe that the API requests to fetch stale and stale_warning counts have `filter[system_profile][host_type]=nil` applied

Note, that the number of such requests is more than they should be. This is not caused by this PR, investigating why. Will improve this once I find the cause


computercamplove:
PR is tested locally - works as expected

Fewwy (approved):

PR #480, opened 2024-04-16
chore: Upgrade eslint-related packages, set up RTL

gkarat:
This is a quick PR to bump linter packages, support RTL and fix a few linter errors appeared after the upgrade in Dashboard.


src/App.js
```diff
@@ -34,7 +34,7 @@ const App = (props) => {
         chrome?.globalFilterScope?.('insights');
         if (chrome?.globalFilterScope) {
             chrome.on('GLOBAL_FILTER_UPDATE', ({ data }) => {
-                const [workloads, SID, selectedTags] = chrome?.mapGlobalFilter?.(data, false, true);
+                const [workloads, SID, selectedTags] = chrome.mapGlobalFilter(data, false, true);
```
bastilian, line 37: I'm just curious, why this change?

gkarat: @bastilian, because ESLint was erroring on this line: optional chaining here still will break/has a huge potential to break things later when the spread items like workloads, SID, etc. are used.

gkarat: I have checked with a few other apps where we use this chrome function, and I haven't seen optional chaining, so I assume we are safe here.

bastilian (approved):
LGTM! Thank you @gkarat!

PR #485, opened 2024-05-15
fix(RHINENG-9846): do not show zero-state while loading data

mkholjuraev:
This makes sure the Zero state banner is not shown during API requests. It also removes redux deprecated function `batch` as now React 18 defaults to batch state updates.

To test:
1. run the application
2. navigate to Dashboard page, observe that no zero state is shown even for milli-seconds if you have systems registered
3. navigate to some other app, and come back no zero state should be displayed
4. open some account without systems registered, observe that now zero state is shown

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,6 @@
 import './App.scss';
 
-import React, { createContext, useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useDispatch } from 'react-redux';
 import { setSIDs, setSelectedTags, setWorkloads } from './AppActions';
 
@@ -9,30 +9,16 @@ import { INVENTORY_TOTAL_FETCH_URL } from './AppConstants';
 import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
 import PageLoading from './PresentationalComponents/PageLoading/PageLoading';
 import { DashboardRoutes } from './DashboardRoutes';
-import ZeroState from './PresentationalComponents/ZeroState/ZeroState';
-
-export const PermissionContext = createContext();
+import ZeroStateWrapper from './PresentationalComponents/ZeroState/AppZeroState';
+import PermissionsProvider from './PresentationalComponents/PermissionsProvider/PermissionsProvider';
 
 const App = (props) => {
     const chrome = useChrome();
     const dispatch = useDispatch();
-    const [permissions, setPermissions] = useState({
-        customPolicies: false,
-        compliance: false,
-        drift: false,
-        advisor: false,
-        remediations: false,
-        patch: false,
-        vulnerability: false,
-        subscriptions: false,
-        ros: false,
-        notifications: false
-    });
-    const [arePermissionsReady, setArePermissionReady] = useState(false);
-    const [hasSystems, setHasSystems] = useState();
-    const [systemsLoading, setSystemsLoading] = useState();
+    const [hasSystems, setHasSystems] = useState(false);
+    const [systemsLoading, setSystemsLoading] = useState(true);
 
-    async function initChrome() {
+    useEffect(() => {
         chrome?.globalFilterScope?.('insights');
         if (chrome?.globalFilterScope) {
             chrome.on('GLOBAL_FILTER_UPDATE', ({ data }) => {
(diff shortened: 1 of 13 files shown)


src/App.js
```diff
@@ -29,17 +30,17 @@ const App = (props) => {
     });
     const [arePermissionsReady, setArePermissionReady] = useState(false);
     const [hasSystems, setHasSystems] = useState();
+    const [hasSystemsReady, setHasSystemsReady] = useState();
```
bastilian, line 33: Isn't this redundant? Wouldn't checking if `hasSystems` is set be the same?

bastilian: Oh, this is similar to what we do with `isLoading` in other places. Maybe `isLoading` would be a better name for it.

src/App.js
```diff
@@ -80,32 +81,39 @@ const App = (props) => {
...
     }, []);
 
+    if (!arePermissionsReady || !hasSystemsReady) {
+        return <PageLoading />;
+    }
+
+    if (hasSystemsReady && !hasSystems) {
+        return <ZeroState />;
+    }
+
```
bastilian, line 100: We should avoid multiple returns.

mkholjuraev: What would be the reason? Is not this cleaner?

src/App.js
mkholjuraev: The reason this is needed is that we should be able differentiate the cases when there no registered systems and when the API request is being made.

src/App.js
bastilian: The reason is that these if statements can get lost in code and usually developers look for the last return at the bottom.

It's also unexpected, as most components only have one return.

mkholjuraev: I would prefer to have if statements in case, the return block has more than 3 condition factors. Because, in some apps, we have so complex conditionally returned component render blocks that it is hard to understand what condition renders what. Also, if we make a change to those conditions it is very easy to do a regression.

src/App.js
gkarat: I would also vote for "isLoading", "systemsLoading" or similar naming. Can we change that?

Also, please set it to some default boolean value so that we strictly distinguish between true/false only.

src/App.js
```diff
@@ -80,32 +81,39 @@ const App = (props) => {
         } catch (e) {
             chrome.hideGlobalFilter();
         }
+
+        setHasSystemsReady(true);
     }
 
     useEffect(() => {
         initChrome();
```
gkarat, line 89: nit-pick: it's an almost redundant function and also named badly (given how much logic we have contained within it not really related to initialising chrome). We would need to refactor it in the soonest future ideally.

src/App.js
```diff
@@ -80,32 +81,39 @@ const App = (props) => {
...
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    if (!arePermissionsReady || !hasSystemsReady) {
+        return <PageLoading />;
+    }
+
+    if (hasSystemsReady && !hasSystems) {
```
gkarat, line 97: hasSystemsReady here is redundant because it will be always true if interpreter reaches this code.

gkarat: ^ could be a reason why I find Sebastian's pick on the multiple return's making sense. I would rather nest conditions under one return. But both options are OK for me.

src/App.js
bastilian: If the function body would only be a handful of lines or even less, I would prefer multiple returns as well, but here it can be easy to scroll past and over look the loading and zero state and recognise them as a possible returned rendered state.

src/App.js
```diff
@@ -80,32 +81,39 @@ const App = (props) => {
...
-                    vulnerability: permissions.vulnerability,
-                    subscriptions: permissions.subscriptions,
-                    ros: permissions.ros,
-                    notifications: permissions.notifications,
-                    hasSystems
-                } }>
-                <DashboardRoutes childProps={ props } />
-            </PermissionContext.Provider>
-            : <PageLoading />
+        <PermissionContext.Provider
```
bastilian, line 102: It would make sense to extract this into a `PermissionsProvider`, with the `permissions` state (and what sets it).

src/App.js
```diff
@@ -80,32 +81,39 @@ const App = (props) => {
...
         initChrome();
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    if (!arePermissionsReady || !hasSystemsReady) {
+        return <PageLoading />;
+    }
+
+    if (hasSystemsReady && !hasSystems) {
+        return <ZeroState />;
```
bastilian, line 98: The `AppZeroState` component has a `hasSystems` prop and a function and allows to wrap children and depending on that it shows a loading state or the children. It would also separate the concerns of the `App` and `ZeroState` component better.

src/App.js
bastilian: Take a look at the comments below. 

We should separate this better, in separate components. The `hasSystems` and the `permissions` don't necessarily belong into the `App` component. If the `ZeroState` (or wrapper component), would contain the `hasSystems` state and render the children or the loading state depending on that, and we had a `PermissionsProvider` component doing something similar, everything would be neatly separated, there wouldn't be multiple returns and no complexity.

src/App.js
mkholjuraev: Nice idea, we need to migrate to the new implementation indeed. Will make the changes and push soon

mkholjuraev: I have found out that the new implementation does not include the changes necessary for the dashboard app. This maybe because the dashboard zero-state is very different from other standard zero-states.

bastilian: @mkholjuraev We can still just build a similar wrapper component, or?

mkholjuraev: I was able to workaround this ^

mkholjuraev:
@RedHatInsights/team-interact this PR is ready for review and would increase dashboard UX greatly. PTAL

bastilian (approved):
Nice! The loading behaviour is now much nicer! Thank you @mkholjuraev!

PR #487, opened 2024-05-29
feat(RHINENG-7048): fec config conversion

Fewwy:
https://issues.redhat.com/browse/RHINENG-7048


fec.config.js
```diff
@@ -0,0 +1,28 @@
...
+    plugins: [],
+    ...(process.env.port ? { port: parseInt(process.env.port) } : {}),
+    moduleFederation: {
+        shared: [
+            {
+                'react-router-dom': {
+                    singleton: true,
+                    import: false,
+                    version: '^6.3.0',
+                },
```
bastilian, line 17: We can grab the version definition from the package.json, then we don't need to manually update it here, if we don't forget. 

We should also add further version requirements like here: https://github.com/RedHatInsights/compliance-frontend/blob/master/fec.config.js#L23-L28

Fewwy: Wooooooooow! I am adding this! :D thanks

src/SmartComponents/Vulnerability/VulnerabilityCard.js
```diff
@@ -20,7 +20,7 @@ import FailState from '../../PresentationalComponents/FailState/FailState';
 import InfoIcon from '../../Icons/InfoIcon';
 import Loading from '../../PresentationalComponents/Loading/Loading';
 import { PieChart } from '../../ChartTemplates/PieChart/PieChartTemplate';
-import global_disabled_color_100 from '@patternfly/react-tokens/dist/esm/global_disabled_color_100';
+import global_disabled_color_100 from '@patternfly/react-tokens';
```
bastilian, line 23: This import is missing the `{..}`

Fewwy: Ok, going to fix that right now

fec.config.js
```diff
@@ -0,0 +1,30 @@
...
+            },
+        ],
+    },
+    exposes: {
+        './AppZeroState': resolve(__dirname, '../src/PresentationalComponents/ZeroState/AppZeroState.js'),
+        './ZeroStateBanner': resolve(__dirname, '../src/PresentationalComponents/ZeroState/ZeroStateBanner.js'),
+        './AppSection': resolve(__dirname, '../src/PresentationalComponents/ZeroState/AppSection.js'),
+        './ZeroStateFooter': resolve(__dirname, '../src/PresentationalComponents/ZeroState/ZeroStateFooter.js'),
+        './RootApp': resolve(__dirname, '../src/AppEntry.js')
+    },
```
bastilian, line 29: This needs to be within `moduleFederation`.

src/SmartComponents/Drift/DriftCard.js
```diff
@@ -136,7 +135,7 @@ const DriftCard = () => {
                                                             dataListCells={[
                                                                 <React.Fragment key={index}>
                                                                     <DataListCell key={`title-${index}`}
-                                                                        wrapModifier={DataListWrapModifier.truncate}>
+                                                                        wrapModifier={'truncate'}>
```
bastilian, line 138: We still should use the `DataListWrapModifier` constant for this we'll need to add an alias for it in the `tsc-transforms` here: https://github.com/RedHatInsights/frontend-components/blob/master/packages/tsc-transform-imports/src/index.ts#L85

Fewwy: I understand that we would like to add more stuff to the cache and use it everywhere we need. But this is the only place in red hat insights org where we use DataListWrapModifier. I think it is safer and faster to change it to a string
image

fec.config.js
Fewwy: If I move the exposed modules to the moduleFederation I get errors like these:
Module not found: Error: Can't resolve '/home/avoznese/RH/src/PresentationalComponents/ZeroState/AppZeroState.js' in '/home/avoznese/RH/insights-dashboard'

src/SmartComponents/Drift/DriftCard.js
bastilian: You just need to add 
```js
DataListWrapModifier: getModuleExplicitLocation(CORE_DIRECTORIES, 'components/DataList'),
``` 

to [packages/tsc-transform-imports/src/index.ts#L85](https://github.com/RedHatInsights/frontend-components/blob/master/packages/tsc-transform-imports/src/index.ts#L85) and bump the package here after it is released.

bastilian: If other apps then also use the `DataListWrapModifier` they won't encounter this issue and worry about replacing it again. It's not only about updating a single app and just get it working, but do it correctly as well and not have to go over everything again.

Fewwy: Ok, understand. I will prepare a PR

Fewwy:
@bastilian I encountered an error with Patch when I opened the link in the dashboard. I think we need to fix this in the patch
https://stage.foo.redhat.com:1337/beta/insights/patch/advisories?offset=0&filter%5Badvisory_type_name%5D=security

Stage works fine
https://console.stage.redhat.com/insights/patch/advisories?offset=0&filter%5Badvisory_type_name%5D=security

image

Fewwy:
@bastilian Another link that now causes an error (it is fine in stage) is "known exploits" for CVE. When I open it I get an error from Vulnerability
image

bastilian (approved):
LGTM! Thank you @Fewwy!

PR #488, opened 2024-06-12
fix(RHINENG-10311): fixing global tags that breaks inventory table

Fewwy:


src/Utilities/useBatchInventoryFetch.js
```diff
@@ -14,14 +14,17 @@ export const useBatchInventoryFetch = (workloads, SID, selectedTags, hasAccess)
...
+        const options = { ...globalFilters(workloads, SID) };
+        const tags = `${selectedTags?.length > 0 ? `&tags=${selectedTags?.join('&tags=')}` : ''}`;
         mounted.current = true;
         const fetchData = async () => {
             try {
                 setIsLoading(true);
-                const inventorySum = options !== 'undefined' && await axios.get(`${INVENTORY_FETCH_URL}`, { params: { ...options } });
-                const inventoryWarningSum = options !== 'undefined' && await axios.get(`${INVENTORY_WARNING_FETCH_URL}`, { params: { ...options } });
-                const inventoryStaleSum = options !== 'undefined' && await axios.get(`${INVENTORY_STALE_FETCH_URL}`, { params: { ...options } });
+                const inventorySum = options !== 'undefined' && await axios.get(`${INVENTORY_FETCH_URL}${tags}`, { params: { ...options } });
```
bastilian, line 23: I know this was here before, but... let's remove the `options !== 'undefined'`, as options are always going to be defined above on line 17. And we can change `{ params: { ...options } }` to `{ params: options }`. In fact, we could rename `options` to `param` and just pass `{ params }` as the second argument of each call.

Fewwy: I updated the code accordingly. Thanks for your advice

src/Utilities/useBatchInventoryFetch.js
```diff
@@ -13,15 +13,17 @@ export const useBatchInventoryFetch = (workloads, SID, selectedTags, hasAccess)
     const [error, setError] = useState(false);
 
     useEffect(() => {
-        //add options to url
-        const options = { ...globalFilters(workloads, SID), ...selectedTags?.length > 0 && { tags: selectedTags } };
+        const params = { ...globalFilters(workloads, SID) };
```
mkholjuraev, line 16:
```suggestion
        const params = globalFilters(workloads, SID);
```

mkholjuraev (approved):
LGTM! The issue with the inventory is fixed. The only issue I see is that if some global filter is applied, really a lot of API calls are made. But most probably this is not related to your PR. Thus approving.

PR #489, opened 2024-06-21
fix(App): Improve loading behaviour

bastilian:
The store appears to change every time the App is rendered this will prevent the store from changing in these cases.

The store changing caused multiple unnecessary rerenders and duplicate requests to APIs, when changing the global filter. This will address the issue.

**How to test**:

1) Open the dashboard
2) Add and remove global filter(s)
3) Observe the number of requests and compare with the number of request on actual stage (or master)

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -8,20 +8,6 @@ module.exports = {
         "@babel/plugin-transform-runtime",
         "@babel/plugin-syntax-dynamic-import",
         "@babel/plugin-proposal-object-rest-spread",
-        [
-          'transform-imports',
-          {
-            '@patternfly/react-icons': {
-              transform: (importName) =>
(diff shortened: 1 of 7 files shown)


babel.config.js
```diff
@@ -8,20 +8,6 @@ module.exports = {
         "@babel/plugin-transform-runtime",
         "@babel/plugin-syntax-dynamic-import",
         "@babel/plugin-proposal-object-rest-spread",
-        [
```
adonispuente, line 11: good catch

adonispuente (approved):
Theres something going on with cypress in multiple repos that is most likely unrelated to this, but while testing this reduced the number of calls by about 50, great job! LGTM!

PR #493, opened 2024-07-26
Update README.md and remove old Travis deploy files

leSamo:


gkarat (commented):
We also can dump this https://github.com/RedHatInsights/insights-dashboard/blob/master/.travis.yml#L15-L16

leSamo:
@gkarat true, I removed the deploy step and upgraded to Node v18 as well

gkarat (approved):
LGTM, thanks @leSamo

PR #494, opened 2024-08-06
fix: Do not hide top buttons behind loading state

leSamo:
https://issues.redhat.com/browse/RHINENG-10939

The loading state now resembles the loaded version of the top bar and allows buttons to be clicked even before the Dashboard is loaded.

### Before
Screenshot from 2024-08-06 17-11-49

### After
Screenshot from 2024-08-06 17-11-01

### Loaded
Screenshot from 2024-08-06 17-11-11


gkarat (approved):
Thanks @leSamo

PR #498, opened 2024-08-09
Fix CentOS and Drift cards collapse not persistent

leSamo:
All cards except Drift and CentOS cards remembered whether they were expanded or collapsed when refreshing the page, except CentOS and Drift, which were always expanded on page load -- this PR fixes the issue.


src/SmartComponents/CentOs/CentOsCard.js
```diff
@@ -44,6 +44,8 @@ const CentOsCard = () => {
         <ExpandableCardTemplate
             appName="tasks"
             className="centos-warning insd-m-toggle-right-on-md"
+            isExpanded={JSON.parse(localStorage.getItem('dashboard_expanded_centos') || 'true')}
```
gkarat, line 47: Can we also put the isExpanded to a component state? This way the CentOsCard component will re-render on the expanded/collapsed change as well.

leSamo: Actually, the way it is currently on CentOS card is consistent with all other cards. The Drift card is the one that is inconsistent. It's inconsistent due to the need to use `isCardExpanded` value at one more place and that is a few lines below to determine whether the Dropdown should be visible.

johnsonm325 (approved):

PR #500, opened 2024-08-09
Fix ROS cards number links being inconsistent

leSamo:
Previously, you could only click the text label under numbers in Resource optimization card to navigate to the ROS app. Now you can also click the numbers which is consistent with Advisor card. 

The change is just moving `InsightsLink` outside of the `Flex`.


AsToNlele (approved):
LGTM 😄

PR #505, opened 2024-09-03
fix(vuln card): Add advisory and affecting filter to Vulnerability links

leSamo:
https://issues.redhat.com/browse/RHINENG-12218


src/SmartComponents/Vulnerability/VulnerabilityCard.js
```diff
@@ -124,7 +124,7 @@ const VulnerabilityCard = () => {
             data: bySeverity?.['0to3.9']
         }];
 
-        const legendLink = (from, to) => `/cves?cvss_from=${from}&cvss_to=${to}`;
+        const legendLink = (from, to) => `/cves?cvss_from=${from}&cvss_to=${to}&advisory_available=true&affecting=edge%2Crpmdnf`;
```
mkholjuraev, line 127: affecting filter behaves different for accounts that have edge devices (immutable systems). Thus, we need to check if an edge device exists in an account and then apply corresponding filter value . For this, we can use the above value: https://github.com/RedHatInsights/insights-dashboard/blob/53b761ce86a174250f22aa6cfe8803b2f4bb541f/src/SmartComponents/Vulnerability/VulnerabilityCard.js#L97

```suggestion
        const legendLink = (from, to) => `/cves?cvss_from=${from}&cvss_to=${to}&advisory_available=true&affecting= ${affectingFilterValue}`;
```

mkholjuraev (changes requested):

src/SmartComponents/Vulnerability/VulnerabilityCard.js
leSamo: Thank you, fixed

gkarat (approved):
LGTM, thanks @leSamo

PR #506, opened 2024-09-12
fix(Vulnerability card): Fix race condition leading to invalid affecting filter

leSamo:
https://issues.redhat.com/browse/RHINENG-12218

### DIFF
--- a/src/SmartComponents/Vulnerability/VulnerabilityCard.js
+++ b/src/SmartComponents/Vulnerability/VulnerabilityCard.js
@@ -162,7 +162,7 @@ const VulnerabilityCard = () => {
         }]);
 
         return [chartData, rows];
-    }, [vulnerabilitiesFetchStatus, vulnerabilities]);
+    }, [vulnerabilitiesFetchStatus, vulnerabilities, affectingFilterValue]);
 
     if (vulnerabilitiesFetchStatus === 'rejected') {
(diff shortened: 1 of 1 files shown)


mkholjuraev:
@leSamo can you provide more info on how to test the PR?

src/SmartComponents/Vulnerability/VulnerabilityCard.js
```diff
@@ -162,7 +162,7 @@ const VulnerabilityCard = () => {
         }]);
 
         return [chartData, rows];
-    }, [vulnerabilitiesFetchStatus, vulnerabilities]);
+    }, [vulnerabilitiesFetchStatus, vulnerabilities, affectingFilterValue]);
```
bastilian, line 165: Hm. Why is this needed as a dependency if it is not used in the memo?

leSamo:
@mkholjuraev this commit fixes the issue mentioned by Max in the Jira ticket comment. Before this commit the link sometimes has the incorrect filter applied (`affecting=true` when edge systems are present, which caused error in Vulnerability when the link is clicked). To test this I would say it's sufficient to refresh the Dashboard a few times on an account with edge systems and check if all of the links in Vulnerability card contain `affecting=rpmdnf,edge`

mkholjuraev (approved):
LGTM!

PR #507, opened 2024-09-26
fix(RHINENG-8056): Advisor read permissions

johnsonm325:
The new `advisor:*:read` permission needs to be accounted for in the permissions provider.


adonispuente (dismissed):
LGTM

adonispuente (approved):
LGTM again

PR #514, opened 2024-10-03
chore(RHINENG-10336): Update cypress code-cov package

johnsonm325:


src/PresentationalComponents/ZeroState/zeroStateConstants.js
```diff
@@ -139,7 +139,7 @@ const COMPLIANCE_ZERO_STATE = {
 
 const INSIGHTS_ZERO_STATE = {
     header: {
-        description: messages.insightsDescription,
+        description: 'Single, consistent management solution across on-premise, hybrid cloud, and public cloud. Included with Red Hat Enterprise Linux subscription. Single, consistent management solution across on-premise, hybrid cloud, and public cloud.',
```
LightOfHeaven1994, line 142: Since we have a constant, I think we need to use that. But instead of `messages.insightsDescription` we need to use `messages.defaultMessage` from here https://github.com/RedHatInsights/insights-dashboard/blob/master/src/Messages.js#L456

src/PresentationalComponents/ZeroState/AppZeroState.test.js
```diff
@@ -14,7 +14,14 @@ jest.mock('@redhat-cloud-services/frontend-components-utilities/interceptors', (
     }))
 }));
 
-const appNames = Object.keys(zeroStateAppList).map(key => key.split('_')[0]);
+const appNames = Object.keys(zeroStateAppList).map(key => {
+    let app = key.split('_')[0];
+    if (app === 'CONTENT' || app === 'RESOURCE') {
+        app = `${app}_${key.split('_')[1]}`;
+    }
```
LightOfHeaven1994, line 21: I think we can remove `if` condition and just have something like `let app = key.replace('_ZERO_STATE', '');`

LightOfHeaven1994 (commented):
Generally looks fine, but there are couple suggestions

src/PresentationalComponents/ZeroState/zeroStateConstants.js
johnsonm325: So, here's what I'll do; those message constants are set up for `intl` which we are doing a soft deprecation on. I'll recreate that constant in a different format.

src/PresentationalComponents/ZeroState/AppZeroState.test.js
johnsonm325: Love this fix!!!

LightOfHeaven1994 (approved):
Amazing! LGTM

PR #524, opened 2024-10-31
fix(ZeroState): Space items evenly

marusak:
Rather than grouping them in center, distribute them evenly. Parent object has max-height set up so it will never get too wild.

Before:
Screenshot from 2024-10-31 18-38-14
After:
Screenshot from 2024-10-31 18-38-55

_(Note: Repositories are already doing this, this is where the inspiration comes from)_


LightOfHeaven1994 (approved):
Looks good, thank you!

PR #580, opened 2025-02-24
fix cypress coverage and collection

opacut:
Fix code coverage scripts and collection so that we have a cypress report generated for codecov.

Review steps:
- npm i
- npm run travis:verify
- check that the coverage-cypress/lcov.info is not empty
- check that the coverage-jest/lcov.info is not empty
- check that .nyc_output/out.json is not empty


LightOfHeaven1994 (approved):
LGTM, but linting errors are valid @opacut

johnsonm325:
I don't know if these failed checks are of concern. I think @LightOfHeaven1994 would know better than me, and he approved. So I say go for it.

PR #592, opened 2025-03-07
Add Cypress coverage to Advisor card

opacut:
Related to the card [RHINENG-16630](https://issues.redhat.com/browse/RHINENG-16630)

### DIFF
--- a/src/SmartComponents/Advisor/Advisor.cy.js
+++ b/src/SmartComponents/Advisor/Advisor.cy.js
@@ -14,6 +14,15 @@ const breakpointColumnsObj = {
     default: 2,
     992: 1
 };
+const CRITICAL_RECOMMENDATIONS_NUM = 3;
+const IMPORTANT_RECOMMENDATIONS_NUM = 16;
+const MODERATE_RECOMMENDATIONS_NUM = 10;
+const LOW_RECOMMENDATIONS_NUM = 0;
+const INCIDENTS_DETECTED_NUM = 4;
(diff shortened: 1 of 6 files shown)


src/SmartComponents/Advisor/Advisor.cy.js
```diff
@@ -0,0 +1,146 @@
...
+
+    it('The count is correct', () => {
+        // Advisor recommendations section
+        cy.get('span:contains("incidents detected")').invoke('text').should((text) => {
+            expect(parseInt(text.split(' ')[0].trim(), 10)).to.equal(4);
+        });
+
+        // Recommendations by total risk section
+        cy.get('a:contains("Critical")').find('span').first().invoke('text').should((text) => {
+            expect(parseInt(text.trim(), 10)).to.equal(3);
```
LightOfHeaven1994, line 52: @opacut I think it would be more readable if we have some constants used here `equal(3)` instead of numbers. Also we can replace this complex logic of parseInt, trim, etc with:

```
        cy.get('a:contains("Critical")')
        .find('span')
        .first()
        .should('have.text', criticalRecommendationNum);
```
and `criticalRecommendationNum` can be obtained from fixture response body

LightOfHeaven1994 (approved):
Generally looks good, but for better readability I would suggest change `The count is correct` test a bit

LightOfHeaven1994:
Looks great, feel free to merge 👍

PR #596, opened 2025-03-26
fix(RHCLOUD-38550): Add ITless env to change button link for ITless envs

aferd:
adding ITless env to change button link for ITless envs. 
[RHCLOUD-38550](https://issues.redhat.com/browse/RHCLOUD-38550)


src/PresentationalComponents/ZeroState/ZeroState.js
```diff
@@ -30,6 +30,8 @@ import InsightsLink from '@redhat-cloud-services/frontend-components/InsightsLin
 
 const ZeroState = () => {
     const intl = useIntl();
+    const getEnvironment = useChrome();
+    const isItLessEnv = getEnvironment() === 'ephem' || 'int' || 'frhStage' || 'frh';
```
gkarat, line 34:
```suggestion
    const isItLessEnv = ['ephem', 'int', 'frhStage', 'frh'].includes(getEnvironment());
```

gkarat, line 34: Also, why it is required to change the button URL for ephemeral environments?

aferd: sorry about the lag here, was out for a week but I switched to using a FF instead.

adonispuente:
Hey, I dont believe this has the desired change youre looking for. The ticket associated with this sort of sounds like its to change some link in inventorys zero state.

The file you altered should change a link in dashboards zero state

aferd:
> Hey, I dont believe this has the desired change youre looking for. The ticket associated with this sort of sounds like its to change some link in inventorys zero state.
> 
> The file you altered should change a link in dashboards zero state

yeah, I searched in inventory-ui as I am targeting this component (see img) which i found to be this [https://github.com/RedHatInsights/insights-dashboard/blob/3fc43f9cf7255cef1ddec753[…]d6d24/src/PresentationalComponents/ZeroState](https://github.com/RedHatInsights/insights-dashboard/blob/3fc43f9cf7255cef1ddec75327917a68681d6d24/src/PresentationalComponents/ZeroState/ZeroStateBanner.js#L23) and the api call  GET https://console.int.openshiftusgov.com/apps/dashboard/js/exposed-./AppZeroState.d1a3f28ca87fb31eaa9d.js comes from insights-dashboard, so I figured this would be the right spot. If not you, could you point me to someone I could double-check this with?

karelhala (approved):

PR #622, opened 2025-05-08
fix(RHINENG-17866): Fix buttons on loading

johnsonm325:
This PR fixes an issue where the "View CVEs" and "View known exploits" buttons don't have the filters they need because the data hasn't fully loaded. The `affectingFilterValue` is set to 'true' during the loading state, and if you click the buttons before the card is loaded, it will cause an error on the vulnerabilities tables.

This PR changes the `affectingFilterValue` to '' on loading, and more importantly, converts the buttons to a Skeleton on loading state so they cannot be clicked.

To test, go to `/insights/dashboard` and see that the "View CVEs" and "View known exploits" buttons are Skeletons on loading. Also, test that you cannot click them.

### DIFF
--- a/src/SmartComponents/Vulnerability/VulnerabilityCard.js
+++ b/src/SmartComponents/Vulnerability/VulnerabilityCard.js
@@ -220,16 +220,19 @@ const VulnerabilityCard = () => {
                                             </p>
                                         </TextContent>
                                     </div>
-                                    <InsightsLink
-                                        aria-label='Security rules link'
-                                        app='vulnerability'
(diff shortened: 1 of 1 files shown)


adonispuente (approved):
LGTM

LightOfHeaven1994:
@johnsonm325 I think we can close this one since it has been fixed here https://github.com/RedHatInsights/insights-dashboard/pull/630

leSamo:
Sorry, I did not notice that this PR when creating mine; I learnt that this bug also happens with other links within Vulnerability card, so I fixed those as well

PR #758, opened 2026-04-22
fix(Kessel): add Kessel checks for Inventory

LightOfHeaven1994:
When I was implementing kessel I wrongly looked at inventory component. so now implementing it proper way
if kessel enabled - we check if we have an access to at least 1 host (relying on backend for permissions).

### How to test:

1. Login as `comp-user-3` as it has kessel FF enabled
2. Navigate to dashboard and see that Inventory component is displayed (before the fix you would see that no permissions)


computercamplove (approved):
LGTM!

PR #759, opened 2026-04-22
chore: add module block with analytics and supportCaseData to deploy config

dajohnso:
## Summary
- Adds the entire `module` block to `frontend.yaml` including analytics keys, manifestLocation, supportCaseData config, and module routes
- Aligns with the pattern used by other frontends (inventory, advisor, tasks, ocp-advisor)

## Test plan
- [ ] Verify the YAML is valid and the deploy CRD parses correctly
- [ ] Confirm analytics events are sent in stage with the correct API key


mtclinton:
@dajohnso think you might need to rebase this one? Claude sometimes has issues with that, so you might have to manually update branch to continue?

mtclinton (changes requested):
See above, PR needs a rebase :)

LightOfHeaven1994 (approved):

PR #761, opened 2026-04-23
fix(RHINENG-26010): follow up for Darkmode

LightOfHeaven1994:
Small follow up (I forgot about this UI content). Fixing link colors for darkmode and also fix wording

Before
image

After
image


Siasurai (approved):
lgtm :relaxed:

PR #764, opened 2026-04-29
fix(RHINENG-26137): deal with permissions correctly based on FF

LightOfHeaven1994:
wait when feature flag is loaded and then fetch permissions


Siasurai (approved):
lgtm :relaxed:

PR #824, opened 2026-07-03
fix: Remove unused commands

swadeley:
Content Sources ZeroState does not display any commands

image


LightOfHeaven1994 (changes requested):
@swadeley it's used, if you click on Blue button that says `Register your system` you will see the content. It's used for Content -> (Advisories, Packages and Systems) that is managed by [patchman-ui](https://github.com/RedHatInsights/patchman-ui) repo

Image

with this change page is going to crash, tested locally

PR #830, opened 2026-07-10
fix: Add RHEL 10 to list of releases

swadeley:
Add RHEL 10 to the list of releases that can be used with the given instructions.


LightOfHeaven1994 (approved):
LGTM!

PR #831, opened 2026-07-10
fix: Add RHEL 10 to list of releases for all

swadeley:
Add RHEL 10 to list of releases for all
This time for all components or plugins.
Also remove unnecessary mention of versions for pre-upgrade analysis.


LightOfHeaven1994 (approved):
LGTM, thank you!

